PR #49, opened 2025-01-25
AAP-39487 Organize tasks in registry, complete demo to test conversions

AlanCoding:
Connect https://github.com/ansible/dispatcher/issues/42, never even got around to a class for task calls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,77 @@ https://github.com/ansible/awx/tree/devel/awx/main/dispatch
 As a part of doing the split, we also want to resolve a number of
 long-standing design and sustainability issues, thus, asyncio.
 
+### Usage
+
+You have a postgres server configured and a python project.
+You will use dispatcher to trigger a background task over pg_notify.
+
+Both your *background dispatcher service* and your *task publisher* process must have
+python configured so that your task is importable.
+
+#### Library
+
+The dispatcher `@task()` decorator is used to register tasks.
+
+See the `tools/test_methods.py` module.
+This defines a dispatcher task and the pg_notify channel it will be sent over.
+
+```python
+from dispatcher.publish import task
+
+@task(queue='test_channel')
+def print_hello():
+    print('hello world!!')
+```
+
+#### Dispatcher service
+
+The dispatcher service needs to be running before you submit tasks.
+This does not make any attempts at message durability or confirmation.
+
+There are 2 ways to run the dispatcher:
+ - Importing and running
+ - A CLI entrypoint `dispatcher-standalone` for demo purposes
+
+The dispatcher ultimately requires code hooks anyway,
+so any convienience the CLI offers is not very relevant
+for running things in production.
+
+Configuration must be provided so it knows how to connect to postgres,
+and what channel to listen to. The demo has this in `dispatcher.yml`.
+When importing and running, you just pass your configuration directly.
+
+The demo config contains a list of channels to listen to which includes `test_channel`.
+
+#### Publisher
+
+This assumes you configured python so that `print_hello` is importable
+from the `test_methods` python module.
+This method does not take any args or kwargs, but if it did, you would
+pass those directly as in, `.delay(*args, **kwargs)`.
+
+The following code will submit `print_hello` to run in the background dispatcher service.
+
+```python
+from test_methods import print_hello
+
+print_hello.delay()
+```
+
+Also valid:
+
+```python
+from test_methods import print_hello
+
+print_hello.apply_async(args=[], kwargs={})
+```
+
+The difference is that `apply_async` takes both args and kwargs as kwargs themselves,
+and allows for additional configuration parameters to come after those.
+
+As of writing, this only works if you have a Django connection configured.
+You can manually pass configuration info (as in the demo) for non-Django use.
+
 ### Manual Demo
 
 You need to have 2 terminal tabs open to run this.
--- a/README.md
+++ b/README.md
@@ -38,6 +38,30 @@ This does not make any attempts at message durability or confirmation.
 
 There are 2 ways to run the dispatcher:
  - Importing and running
+  ```python
+  from dispatcher.main import DispatcherMain
+  import asyncio
+  
+  pgnotify_dsn = "dbname=postgres user=postgres"
+  config = {
+      "producers": {
+          "brokers": {
+              "pg_notify": {"conninfo": pgnotify_dsn},
+              "channels": [
+                  "test_channel",
+              ],
+          },
+      },
+      "pool": {"max_workers": 4},
+  }
+  loop = asyncio.get_event_loop()
+  dispatcher = DispatcherMain(config)
+  
+  
+  try:
+      loop.run_until_complete(dispatcher.main())
+  finally:
+      loop.close()
  - A CLI entrypoint `dispatcher-standalone` for demo purposes
 
 The dispatcher ultimately requires code hooks anyway,
--- a/README.md
+++ b/README.md
@@ -35,44 +35,39 @@ def print_hello():
 
 The dispatcher service needs to be running before you submit tasks.
 This does not make any attempts at message durability or confirmation.
+If you submit a task in an outage of the service, it will be dropped.
 
-There are 2 ways to run the dispatcher:
- - Importing and running
-  ```python
-  from dispatcher.main import DispatcherMain
-  import asyncio
-  
-  pgnotify_dsn = "dbname=postgres user=postgres"
-  config = {
-      "producers": {
-          "brokers": {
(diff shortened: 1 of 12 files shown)


dispatcher/publish.py
```diff
@@ -1,19 +1,30 @@
...
+from dispatcher.registry import DispatcherMethodRegistry
+from dispatcher.registry import registry as default_registry
+from dispatcher.utils import DispatcherCallable
 
 logger = logging.getLogger('awx.main.dispatch')
 
 
-def serialize_task(f):
-    return '.'.join([f.__module__, f.__name__])
+def _task_decorator(fn: DispatcherCallable, registry=default_registry, **kwargs) -> DispatcherCallable:
```
Alex-Izquierdo, line 11: should we allow or disallow explicitly positional arguments in this function?

Alex-Izquierdo, line 11: The registry should have also type hint, don't think you?

dispatcher/publish.py
```diff
@@ -1,19 +1,30 @@
...
+    dmethod = registry.register(fn, **kwargs)
 
-class task:
+    setattr(fn, 'apply_async', dmethod.apply_async)
+    setattr(fn, 'delay', dmethod.delay)
+
+    return fn
+
+
+def task(
```
Alex-Izquierdo, line 22: The return type in the type hint is missing.

dispatcher/publish.py
```diff
@@ -1,19 +1,30 @@
...
+    fn_maybe: Optional[DispatcherCallable] = None,
+    queue: Optional[str] = None,
+    on_duplicate: Optional[str] = None,
```
Alex-Izquierdo, line 25: Can we document the meaning of the arguments? Especially `on_duplicate`

dispatcher/registry.py
```diff
@@ -0,0 +1,146 @@
...
+
+        return self.fn
+
+    def publication_defaults(self) -> dict:
+        defaults = {'task': self.serialize_task(), 'time_pub': time.time()}
+        if self.on_duplicate not in [DuplicateBehavior.parallel.value, None]:
+            defaults['on_duplicate'] = self.on_duplicate
+        return defaults
+
+    def delay(self, *args, **kwargs):
```
Alex-Izquierdo, line 54: Return value missing in type hint

dispatcher/registry.py
```diff
@@ -0,0 +1,146 @@
...
+        return self.apply_async(args, kwargs)
+
+    def get_async_body(self, args=None, kwargs=None, uuid=None, delay=None, **kw):
```
Alex-Izquierdo, line 57: Return value missing in type hint

dispatcher/registry.py
```diff
@@ -0,0 +1,146 @@
...
+
+        if delay is not None:
+            body['delay'] = delay
+
+        # TODO: callback to add other things, guid in case of AWX
+
+        body.update(**kw)
+        return body
+
+    def apply_async(self, args=None, kwargs=None, queue=None, uuid=None, delay=None, connection=None, config=None, on_duplicate=None, **kw):
```
Alex-Izquierdo, line 74: Return value missing in type hint

dispatcher/utils.py
```diff
@@ -1,14 +1,25 @@
...
-from typing import Callable, Optional
+from typing import Callable, Optional, Protocol, Type, Union, runtime_checkable
+
+
+@runtime_checkable
+class RunnableClass(Protocol):
+    def run(self, *args, **kwargs) -> None: ...
+
+
+MODULE_METHOD_DEL = ':'
```
Alex-Izquierdo, line 11: Would you agree to rename it with a more meaningful name? For example `MODULE_METHOD_DELIMITER` ?
I think "DEL" can be ambiguous.

Alex-Izquierdo (commented):
All my comments are so far code style issues. I would like to test this PR to ensure it does not brake what we already have in EDA podman PoC and try understand better the core purpose.

Alex-Izquierdo:
@AlanCoding since it is a draft, Should I try to test it or is it waste because it's WIP?

dispatcher/publish.py
AlanCoding: We _must_ allow `fn` as a positional argument because it's a decorator. I tried to add `, *` but something about that wasn't working. Now I'm not remembering what.

Alex-Izquierdo: Sure `fn`, I did not express myself well. I meant each other pos args (*args) or not supporting the `registry` as positional and/or enforce kwargs after `fn`

AlanCoding: Yes, 100%. We have never used positional kwargs to `@task` other than native decorator functionality. Anything other than the function should be strictly key word only. It seems that adding `*, ` is working for me now, so I'll push a commit with that.

And now that I look at this, I've also convinced myself that only `@task()` should be allowed, not `@task`.

AlanCoding:
> Should I try to test it or is it waste because it's WIP?

We should very much verify with your tests! This might be a good starting place to establish a pattern of running those before merging so we don't undo any progress.

This was originally important to me to "finish" testing, meaning add proper integration tests to include all the features I had previously written into the demo script. However, I found that very challenging to do in isolation without also implementing some elements of https://github.com/ansible/dispatcher/issues/42

It was hard/awkward to finish writing some of that without also having https://github.com/ansible/dispatcher/issues/44, but so-far, I have intentionally excluded that from the scope. So I could mark this as my stopping place, and then work with you to get a pattern for verification before merge. Or, depending on your timeline, I leave leave it open and start on the config work (issue 44), and then test it after.

Alex-Izquierdo:
@AlanCoding So far this PR seems to be working fine with EDA (after changing the syntax of the celerybeat functions, of course)

AlanCoding:
Oh no, I didn't mean to change the contract with the schedule definition. But you're right because of `MODULE_METHOD_DELIMITER`.

I meant to set that back to a `"."` after confirming it worked with another character. Let me push a commit with that change and then I'll take this out of draft. I didn't mean for this to constitute a contract change.

README.md
```diff
@@ -8,6 +8,77 @@ https://github.com/ansible/awx/tree/devel/awx/main/dispatch
...
+    print('hello world!!')
+```
+
+#### Dispatcher service
+
+The dispatcher service needs to be running before you submit tasks.
+This does not make any attempts at message durability or confirmation.
+
+There are 2 ways to run the dispatcher:
+ - Importing and running
```
Alex-Izquierdo, line 40: Lets add a snippet as example to run the process, WDYT?
```suggestion
 - Importing and running
  ```python
  from dispatcher.main import DispatcherMain
  import asyncio
  
  pgnotify_dsn = "dbname=postgres user=postgres"
  config = {
      "producers": {
          "brokers": {
              "pg_notify": {"conninfo": pgnotify_dsn},
              "channels": [
                  "test_channel",
              ],
          },
      },
      "pool": {"max_workers": 4},
  }
  loop = asyncio.get_event_loop()
  dispatcher = DispatcherMain(config)
  
  
  try:
      loop.run_until_complete(dispatcher.main())
  finally:
      loop.close()
  ```
```

dispatcher/utils.py
```diff
@@ -1,14 +1,25 @@
...
 
 
 def resolve_callable(task: str) -> Optional[Callable]:
     """
     Transform a dotted notation task into an imported, callable function, e.g.,
 
-    awx.main.tasks.system.delete_inventory
-    awx.main.tasks.jobs.RunProjectUpdate
+    awx.main.tasks.system:delete_inventory
+    awx.main.tasks.jobs:RunProjectUpdate
```
Alex-Izquierdo, line 22: After change the DELIMITER this needs to be updated again:
```suggestion
    awx.main.tasks.system.delete_inventory
    awx.main.tasks.jobs.RunProjectUpdate
```

README.md
AlanCoding: I was looking over this README and noticed the same thing. It gives code for the definition and publisher, but not the service. So agree.

README.md
```diff
@@ -8,6 +8,101 @@ https://github.com/ansible/awx/tree/devel/awx/main/dispatch
...
+      "pool": {"max_workers": 4},
+  }
+  loop = asyncio.get_event_loop()
+  dispatcher = DispatcherMain(config)
+  
+  
+  try:
+      loop.run_until_complete(dispatcher.main())
+  finally:
+      loop.close()
```
Alex-Izquierdo, line 64: Sorry, I think GH messed it up a little handling nested code blocks, it is missing the closing block
```suggestion
      loop.close()
      ```
```

PR #53, opened 2025-02-07
Support for bind=True kwarg

AlanCoding:
Obviously, still need a test for this.

This is intended to leave room to grow for later. But this will accomplish the goal of getting the `uuid` of the current task, which I've argued the use case to @Alex-Izquierdo and @chrismeyersfsu 

Small patch, but very big usage consequences for clustering issues.


Alex-Izquierdo (commented):
Can we add documentation about the meaning and use of "bind" argument? I feel it is not intuitive. In celery you can find both: user docs and doctrings.

AlanCoding:
I've rebased this now so it resolves conflicts with the `timeout` addition, which is mostly in the same contracts.

Alex-Izquierdo (approved):

PR #64, opened 2025-02-11
A global config and `setup` method

AlanCoding:
Fixes https://github.com/ansible/dispatcher/issues/44

The argument for this is difficult and something I've struggled with. But it's a strong argument. We have 3 different actors:

1. The service itself, it runs tasks
2. The publisher, it submits tasks
3. Any `dispatcherctl` type command, exists in AWX, does control-and-reply to get debugging information

Because these are all invoked separately, we should requires the `setup()` method to be called, and if not, throw an error. Otherwise, it's very difficult to assure that the service is using the same config as the publisher, for example. On higher levels, this is easy to enforce using the Django `.ready()` method. But this will set the tone for how tests are written on the psycopg layer.

### DIFF
--- a/dispatcher.yml
+++ b/dispatcher.yml
@@ -1,20 +1,23 @@
 # Demo config
 ---
-pool:
-  max_workers: 3
-producers:
-  brokers:
-    # List of channels to listen on
+service:
+  max_workers: 4
+brokers:
+  pg_notify:
+    config:
+      conninfo: dbname=dispatch_db user=dispatch password=dispatching host=localhost port=55777
+    sync_connection_factory: dispatcher.brokers.pg_notify.connection_saver
     channels:
     - test_channel
     - test_channel2
     - test_channel3
-    pg_notify:
-      # Database connection details
-      conninfo: dbname=dispatch_db user=dispatch password=dispatching host=localhost
-        port=55777
-  scheduled:
-    'lambda: __import__("time").sleep(1)':
-      schedule: 3
-    'lambda: __import__("time").sleep(2)':
-      schedule: 3
+    default_publish_channel: test_channel
+producers:
+  ScheduledProducer:
+    task_schedule:
+      'lambda: __import__("time").sleep(1)':
+        schedule: 3
+      'lambda: __import__("time").sleep(2)':
+        schedule: 3
+publish:
+  default_broker: pg_notify
--- a/dispatcher.yml
+++ b/dispatcher.yml
@@ -1,5 +1,6 @@
 # Demo config
 ---
+version: 2
 service:
   pool_kwargs:
     max_workers: 4
--- a/dispatcher.yml
+++ b/dispatcher.yml
@@ -1,7 +1,8 @@
 # Demo config
 ---
 service:
-  max_workers: 4
+  pool_kwargs:
+    max_workers: 4
 brokers:
   pg_notify:
     config:
--- a/dispatcher/brokers/base.py
+++ b/dispatcher/brokers/base.py
@@ -1,0 +1,25 @@
+from abc import abstractmethod
+from typing import Optional
+
+
+class BaseBroker:
+    @abstractmethod
+    async def connect(self): ...
+
+    @abstractmethod
+    async def aprocess_notify(self, connected_callback=None): ...
+
+    @abstractmethod
+    async def apublish_message(self, channel: Optional[str] = None, message: str = '') -> None: ...
+
+    @abstractmethod
+    async def aclose(self) -> None: ...
+
+    @abstractmethod
+    def get_connection(self): ...
+
+    @abstractmethod
+    def publish_message(self, channel=None, message=None): ...
+
+    @abstractmethod
+    def close(self): ...
--- a/dispatcher/brokers/base.py
+++ b/dispatcher/brokers/base.py
@@ -2,8 +2,8 @@ from typing import AsyncGenerator, Optional, Protocol
 
 
 class BaseBroker(Protocol):
-    # NOTE: should be async def, but conflicts with lack of yield statement, which implementers would have
-    def aprocess_notify(self, connected_callback=None) -> AsyncGenerator[tuple[str, str], None]: ...
+    async def aprocess_notify(self, connected_callback=None) -> AsyncGenerator[tuple[str, str], None]:
+        yield ('', '')  # yield affects CPython type https://github.com/python/mypy/pull/18422
 
     async def apublish_message(self, channel: Optional[str] = None, message: str = '') -> None: ...
 
--- a/dispatcher/brokers/base.py
+++ b/dispatcher/brokers/base.py
@@ -1,8 +1,9 @@
-from typing import Optional, Protocol
+from typing import AsyncGenerator, Optional, Protocol
 
 
 class BaseBroker(Protocol):
-    async def aprocess_notify(self, connected_callback=None): ...
+    # NOTE: should be async def, but conflicts with lack of yield statement, which implementers would have
+    def aprocess_notify(self, connected_callback=None) -> AsyncGenerator[tuple[str, str], None]: ...
 
     async def apublish_message(self, channel: Optional[str] = None, message: str = '') -> None: ...
 
--- a/dispatcher/brokers/base.py
+++ b/dispatcher/brokers/base.py
@@ -1,25 +1,13 @@
-from abc import abstractmethod
-from typing import Optional
+from typing import Optional, Protocol
 
 
-class BaseBroker:
-    @abstractmethod
-    async def connect(self): ...
-
-    @abstractmethod
+class BaseBroker(Protocol):
     async def aprocess_notify(self, connected_callback=None): ...
 
-    @abstractmethod
     async def apublish_message(self, channel: Optional[str] = None, message: str = '') -> None: ...
 
-    @abstractmethod
     async def aclose(self) -> None: ...
 
-    @abstractmethod
-    def get_connection(self): ...
-
-    @abstractmethod
     def publish_message(self, channel=None, message=None): ...
 
-    @abstractmethod
     def close(self): ...
--- a/dispatcher/brokers/pg_notify.py
+++ b/dispatcher/brokers/pg_notify.py
@@ -16,20 +16,51 @@ Thus, all psycopg-lib-specific actions must happen here.
 """
 
 
-class PGNotifyBase:
+async def acreate_connection(**config) -> psycopg.AsyncConnection:
+    "Create a new asyncio connection"
+    return await psycopg.AsyncConnection.connect(**config)
+
+
+def create_connection(**config) -> psycopg.Connection:
+    return psycopg.Connection.connect(**config)
+
+
+class Broker:
 
     def __init__(
         self,
         config: Optional[dict] = None,
+        async_connection_factory: Optional[str] = None,
+        sync_connection_factory: Optional[str] = None,
+        sync_connection: Optional[psycopg.Connection] = None,
+        async_connection: Optional[psycopg.AsyncConnection] = None,
         channels: Union[tuple, list] = (),
         default_publish_channel: Optional[str] = None,
     ) -> None:
         """
+        config - kwargs to psycopg connect classes, if creating connection this way
+        (a)sync_connection_factory - importable path to callback for creating
+          the psycopg connection object, the normal or synchronous version
+          this will have the config passed as kwargs, if that is also given
+        async_connection - directly pass the async connection object
+        sync_connection - directly pass the async connection object
         channels - listening channels for the service and used for control-and-reply
         default_publish_channel - if not specified on task level or in the submission
           by default messages will be sent to this channel.
           this should be one of the listening channels for messages to be received.
         """
+        if not (config or async_connection_factory or async_connection):
+            raise RuntimeError('Must specify either config or async_connection_factory')
+
+        if not (config or sync_connection_factory or sync_connection):
+            raise RuntimeError('Must specify either config or sync_connection_factory')
+
+        self._async_connection_factory = async_connection_factory
+        self._async_connection = async_connection
+
+        self._sync_connection_factory = sync_connection_factory
+        self._sync_connection = sync_connection
+
         if config:
             self._config: dict = config.copy()
             self._config['autocommit'] = True
@@ -51,52 +82,26 @@ class PGNotifyBase:
 
         raise ValueError('Could not determine a channel to use publish to from settings or PGNotify config')
 
-    def get_connection_method(self, factory_path: Optional[str] = None) -> Callable:
-        "Handles settings, returns a method (async or sync) for getting a new connection"
-        if factory_path:
-            factory = resolve_callable(factory_path)
-            if not factory:
-                raise RuntimeError(f'Could not import connection factory {factory_path}')
-            return factory
-        elif self._config:
-            return self.create_connection
-        else:
-            raise RuntimeError('Could not construct connection for lack of config or factory')
+    # --- asyncio connection methods ---
 
-    def create_connection(self): ...
-
-
-class AsyncBroker(PGNotifyBase):
-    def __init__(
-        self,
-        config: Optional[dict] = None,
-        async_connection_factory: Optional[str] = None,
-        sync_connection_factory: Optional[str] = None,  # noqa
-        connection: Optional[psycopg.AsyncConnection] = None,
-        **kwargs,
-    ) -> None:
-        if not (config or async_connection_factory or connection):
-            raise RuntimeError('Must specify either config or async_connection_factory')
-
-        self._async_connection_factory = async_connection_factory
-        self._connection = connection
-
-        super().__init__(config=config, **kwargs)
-
-    async def get_connection(self) -> psycopg.AsyncConnection:
-        if not self._connection:
-            factory = self.get_connection_method(factory_path=self._async_connection_factory)
-            connection = await factory(**self._config)
-            self._connection = connection
+    async def aget_connection(self) -> psycopg.AsyncConnection:
+        "Return existing connection or create a new one"
+        if not self._async_connection:
+            if self._async_connection_factory:
+                factory = resolve_callable(self._async_connection_factory)
+                if not factory:
+                    raise RuntimeError(f'Could not import async connection factory {self._async_connection_factory}')
+                connection = await factory(**self._config)
+            elif self._config:
+                connection = await acreate_connection(**self._config)
+            else:
+                raise RuntimeError('Could not construct async connection for lack of config or factory')
+            self._async_connection = connection
             return connection  # slightly weird due to MyPY
-        return self._connection
+        return self._async_connection
 
-    @staticmethod
-    async def create_connection(**config) -> psycopg.AsyncConnection:
-        return await psycopg.AsyncConnection.connect(**config)
-
-    async def aprocess_notify(self, connected_callback=None):
-        connection = await self.get_connection()
+    async def aprocess_notify(self, connected_callback: Optional[Callable] = None):  # public
+        connection = await self.aget_connection()
         async with connection.cursor() as cur:
             for channel in self.channels:
                 await cur.execute(f"LISTEN {channel};")
@@ -110,8 +115,8 @@ class AsyncBroker(PGNotifyBase):
                 async for notify in connection.notifies():
                     yield notify.channel, notify.payload
 
-    async def apublish_message(self, channel: Optional[str] = None, message: str = '') -> None:
-        connection = await self.get_connection()
+    async def apublish_message(self, channel: Optional[str] = None, message: str = '') -> None:  # public
+        connection = await self.aget_connection()
         channel = self.get_publish_channel(channel)
 
         async with connection.cursor() as cur:
@@ -123,38 +128,26 @@ class AsyncBroker(PGNotifyBase):
         logger.debug(f'Sent pg_notify message of {len(message)} chars to {channel}')
 
     async def aclose(self) -> None:
-        if self._connection:
-            await self._connection.close()
-            self._connection = None
+        if self._async_connection:
+            await self._async_connection.close()
+            self._async_connection = None
 
-
-class SyncBroker(PGNotifyBase):
-    def __init__(
-        self,
-        config: Optional[dict] = None,
-        async_connection_factory: Optional[str] = None,  # noqa
-        sync_connection_factory: Optional[str] = None,
-        connection: Optional[psycopg.Connection] = None,
-        **kwargs,
-    ) -> None:
-        if not (config or sync_connection_factory or connection):
-            raise RuntimeError('Must specify either config or async_connection_factory')
-
-        self._sync_connection_factory = sync_connection_factory
-        self._connection = connection
-        super().__init__(config=config, **kwargs)
+    # --- synchronous connection methods ---
 
     def get_connection(self) -> psycopg.Connection:
-        if not self._connection:
-            factory = self.get_connection_method(factory_path=self._sync_connection_factory)
-            connection = factory(**self._config)
-            self._connection = connection
+        if not self._sync_connection:
+            if self._sync_connection_factory:
+                factory = resolve_callable(self._sync_connection_factory)
+                if not factory:
+                    raise RuntimeError(f'Could not import connection factory {self._sync_connection_factory}')
+                connection = factory(**self._config)
+            elif self._config:
+                connection = create_connection(**self._config)
+            else:
+                raise RuntimeError('Could not construct connection for lack of config or factory')
+            self._sync_connection = connection
             return connection
-        return self._connection
-
-    @staticmethod
-    def create_connection(**config) -> psycopg.Connection:
-        return psycopg.Connection.connect(**config)
+        return self._sync_connection
 
     def publish_message(self, channel: Optional[str] = None, message: str = '') -> None:
         connection = self.get_connection()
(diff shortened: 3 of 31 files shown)


dispatcher/brokers/base.py
```diff
@@ -0,0 +1,43 @@
+from abc import abstractmethod
+from typing import Any, Iterable, Optional
+
+
+class BaseBroker:
```
Alex-Izquierdo, line 5: Do we want to declare this class as an abstract interface? In such a case, methods should also define return values and probably other interfaces like the returned one by `get_connection` otherwise the implementation would be depending on psycopg regardless of the abstraction.

dispatcher/brokers/pg_notify.py
```diff
@@ -13,87 +16,108 @@
...
-        conn.close()
+class PGNotifyBase(BaseBroker):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self._config:
+            self._config = self._config.copy()
+            self._config['autocommit'] = True
+
+
+class AsyncBroker(PGNotifyBase):
```
Alex-Izquierdo, line 27: Probably I missing something but I see some code duplication when handling the connection between the async broker and the sync broker and I wonder if would make sense to add the logic in PGNotifyBase using polymorphism to delegate connection instantiation and implement `create_connection` as class method or is it something that you have discarded for some reason?

dispatcher/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+        for task_name, options in self.tasks.items():
+            method = resolve_callable(task_name)
+            if method:
+                DispatcherDecorator(task_registry, **options)(method)
+            else:
+                raise RuntimeError(f'Could not locate task given in config: {task_name}')
+
+
+def settings_from_file(path: str) -> DispatcherSettings:
```
Alex-Izquierdo, line 29: should we handle IO and yaml parsing errors here?

dispatcher/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+settings = LazySettings()
+
+
+def setup(config: Optional[dict] = None, file_path: Optional[str] = None):
+    if config:
+        settings._wrapped = DispatcherSettings(config)
+    elif file_path:
+        settings._wrapped = settings_from_file(file_path)
+    settings._wrapped = settings_from_env()
```
Alex-Izquierdo, line 67: have you missed `else` statement before this line?

dispatcher/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+empty = object()
+
+
+class LazySettings:
+    def __init__(self) -> None:
+        self._wrapped: Optional[DispatcherSettings] = None
+
+    def __getattr__(self, name):
+        if self._wrapped is empty:
```
Alex-Izquierdo, line 51: Why not just check if `_wrapped` is none?

dispatcher/config.py
```diff
@@ -0,0 +1,77 @@
...
+def setup(config: Optional[dict] = None, file_path: Optional[str] = None):
+    if config:
+        settings._wrapped = DispatcherSettings(config)
+    elif file_path:
+        settings._wrapped = settings_from_file(file_path)
+    settings._wrapped = settings_from_env()
+
+
+@contextmanager
+def temporary_settings(config):
```
Alex-Izquierdo, line 71: Is this function for testing purposes?

dispatcher/registry.py
```diff
@@ -82,11 +83,16 @@ def apply_async(self, args=None, kwargs=None, queue=None, uuid=None, connection=
 
         obj = self.get_async_body(args=args, kwargs=kwargs, uuid=uuid, **kw)
 
+        from dispatcher.conf import settings
+
+        for broker_name, broker_kwargs in settings.brokers.items():
```
Alex-Izquierdo, line 88: why we would have multiple brokers and why we would take just the first one? I admit I'm lost at this point.

Alex-Izquierdo (commented):
Since this PR is a draft, I'm not sure if some of my questions are premature. I'n sorry in that case.

dispatcher/config.py
```diff
@@ -0,0 +1,77 @@
...
+from dispatcher.utils import resolve_callable
+
+
+class DispatcherSettings:
+    def __init__(self, config: dict, task_registry: DispatcherMethodRegistry = registry) -> None:
+        self.brokers: list = config.get('brokers', [])
+        self.producers: list = config.get('producers', [])
+        self.service: dict = config.get('service', {'max_workers': 3})
+        self.callbacks: dict = config.get('callbacks', {})
+        self.tasks: dict = config.get('tasks', {})
```
Alex-Izquierdo, line 18: is there a real usecase for tasks defined at this level? I wonder if should try to limit ourselves for the strictly needed funcionality.

dispatcher/brokers/pg_notify.py
AlanCoding: We can't share that method b/c one version is `def` and the other version is `async def`. Good eye. I don't think there's a solution to consolidate thta.

dispatcher/config.py
```diff
@@ -0,0 +1,77 @@
...
+from dispatcher.registry import DispatcherMethodRegistry, registry
+from dispatcher.utils import resolve_callable
+
+
+class DispatcherSettings:
+    def __init__(self, config: dict, task_registry: DispatcherMethodRegistry = registry) -> None:
+        self.brokers: list = config.get('brokers', [])
+        self.producers: list = config.get('producers', [])
+        self.service: dict = config.get('service', {'max_workers': 3})
+        self.callbacks: dict = config.get('callbacks', {})
```
Alex-Izquierdo, line 17: “callbacks” would need to satisfy a certain schema? I understand that are callbacks called in a given point of the flow. Should we have callbacks per task? (e.g., pretask, posttask triggers)

dispatcher/brokers/base.py
AlanCoding: Yes, you see I'm already using something from `abc`. This is kind of my intent. Also, much of this was necessary to make `mypy` happy, which I do think helps make better code. However, I don't know, specifically, what else you're thinking of.

dispatcher/brokers/pg_notify.py
Alex-Izquierdo: An async func can call a sync func. 
I mean something like this in PGNotifyBase
```
    async def get_async_connection(self) -> psycopg.AsyncConnection:
    def get_sync_connection(self) -> psycopg.Connection:
    def create_connection(self, factory_attr, create_method):
```
plus:
```
class AsyncBroker(PGNotifyBase):
    async def get_connection(self) -> psycopg.AsyncConnection:
        return await self.get_async_connection()

class SyncBroker(PGNotifyBase):
    def get_connection(self) -> psycopg.Connection:
        return self.get_sync_connection()

```

dispatcher/brokers/base.py
Alex-Izquierdo: My assumption is that you would want to define the interface for the broker, because we should be able to develop other brokers by just satisfying the interface (it does not seem to be the case in the near future). In that case, we should define as well an interface for the connection, for example. I guess we are going to work only with pgnotify for now so get_connection actually relies on the interface of the psycopg connection and we should indicate it in the interface.

I would discourage a class that are both, base class and interface. If the point is just make mypy happy, we can also define a protocol:
https://docs.python.org/3/library/typing.html#typing.Protocol

either one way of another, I would say the get_connection method only supports [PGConn](https://github.com/psycopg/psycopg/blob/be4e811369e446188e3ffd6c000dd32eec594023/psycopg/psycopg/pq/abc.py#L22) for now. 

My point is that in the end, abstract methods should define as well their expected return values. :)

dispatcher/config.py
AlanCoding: oh this is a total goof. I abandoned `empty` and this shouldn't be here. Sorry about that. Little premature of a patch.

dispatcher/config.py
AlanCoding: Writing unit tests against a non-global config would be best practice, and I have some existing tests that use a non-global registry. My thought was that I needed this to test a non-global config _with a non-global registry_.

I agree, it would be best to avoid this. There is an academic reason with the attached `delay` and `apply_async` methods, but more tactically, it would be better to avoid import dependencies. But I need to get the ball rolling on the unit tests, and I won't be confident until then that it'll work out.

dispatcher/config.py
AlanCoding: :+1: Yes. I don't have a pattern in mind for how to do that. Some hooks are sure to be:

 - worker pool, pre-fork
 - worker start
 - worker, before starting task
 - worker, after finishing task

Alex-Izquierdo: Yeah, I wonder if these callbacks should be set in the config. I feel they are more "codeable" matters and maybe it can be set by the implementer through a hook register or something like that, as same as the task callback would be set at the definition of the task.

AlanCoding: I have a nagging concern for any hook / register / code-able thing. You might set it in the service, but that won't set it for the publisher. The logging context data is a good example where you need synchronization between the task submitter and the service running the tasks. So if you registered this stuff, you would have to be sure to do it for both services. That's why I'm going in the direction of a `setup` method that does all those registrations. It will be simpler to funnel all users of the dispatcher through that single setup call (or error if they don't).

I'm still shifting around on some things, like, maybe `setup` can configure task registry & callbacks, which are separate from `settings`.

For now I've booted callbacks out of scope of this. But I've learned some things from this.

dispatcher/registry.py
AlanCoding: This was addressed by adding to the config. My thought process was still evolving when I first pushed this. Now this part should be coherent, although I can't yet get rid of the in-line import.

https://github.com/ansible/dispatcher/blob/a77cbb2aae302e32558a104aed0155bd94945943/dispatcher.yml#L22-L23

dispatcher/config.py
AlanCoding: I envision the dab_task app using this a lot. Because this should be the "easy button" for testing in app space. The app doesn't create brokers or pool objects. It says "here's my config, go run". But for testing, the app may need to change the config out for every test.

For testing in the dispatcher:
 - unit tests shouldn't generally have any need for this, these should directly create objects, and might use dictionaries from inside of the config to help
 - integration tests will create objects from "factories" that I just introduced, but generally should pass the config as a kwarg, instead of using the global context. But honestly it would be fine either way.

I've got tests working, but I'm still looking at smoothing out the standard practice.

dispatcher/config.py
AlanCoding: Is it okay if I want those to be left as exceptions? I worry that's going to be distracting work with no value-add.

Assuming a valid file, and valid YAML, we've got plenty to worry about regarding valid vs. invalid content.

dispatcher/brokers/pg_notify.py
AlanCoding: I didn't believe you at first, but you were right. I wound up with something in this general spirit. The key thing I was missing was that the factory method had to be passed around.

dispatcher/pool.py
```diff
@@ -186,7 +189,12 @@ async def manage_timeout(self) -> None:
             self.events.timeout_event.clear()
 
     async def up(self) -> None:
-        process = self.process_manager.create_process((self.next_worker_id,))
+        process = self.process_manager.create_process(
+            (
+                self.settings_stash,
```
AlanCoding, line 194: We will need this in order to enable use of forkserver for creating worker processes. Because we can't rely on references to global objects from the parent in that model. So we have to pass data that we need.

tests/integration/publish/test_registry.py
```diff
@@ -0,0 +1,28 @@
...
+        # Can not run a method if we do not have a queue
+        with pytest.raises(ValueError):
+            dmethod.apply_async()
+
+        # But providing a queue at time of submission works
+        with mock.patch('dispatcher.brokers.pg_notify.SyncBroker.publish_message') as mock_publish_method:
+            dmethod.apply_async(queue='fooqueue')
+        mock_publish_method.assert_called_once_with(channel='fooqueue', message=mock.ANY)
+
+    mock_publish_method.assert_called_once()
```
AlanCoding, line 28: This was moved over from a unit test, because the point at which this is enforced was pushed back later in the process, to the broker class. This required access to the database, so moved here.

AlanCoding:
Additional incoming change:

I've changed my tune on how `__init__` arguments should be passed. We should be consistent, and this would result in needing to pass _related objects_ as arguments when one object needs another. So the worker pool needs to be passed the process manager. Then the main object needs to be passed the worker pool. Then the factories will work out how to extract the kwargs from the config file. This suggests a new YAML structure like this:

```yaml
service:
  pool_cls: dispatcher.pool.WorkerPool
  pool_kwargs:
    min_workers: 2
    max_workers: 12
  process_manager_cls: dispatcher.process.ForkServerManager
  process_manager_kwargs:
    preload_modules: my_app.hazmat
```

I was trying to make something work where we passed the process-manager kwargs into the main object... and this pattern doesn't scale well. And this all needs to be as "dumb", and straight-to-the-code as possible. Rephrasing, making the config file look messy is fine if it makes things simpler overall.

dispatcher.yml
```diff
@@ -1,20 +1,23 @@
...
-        port=55777
-  scheduled:
-    'lambda: __import__("time").sleep(1)':
-      schedule: 3
-    'lambda: __import__("time").sleep(2)':
-      schedule: 3
+    default_publish_channel: test_channel
+producers:
+  ScheduledProducer:
+    task_schedule:
```
Alex-Izquierdo, line 17: I was wondering that scheduled tasks are usually more dependent on the logic of the application and as any other task, may be more useful if they can be registered in runtime by the application. Is something that you have considered @AlanCoding  ?

Alex-Izquierdo (commented):
Should have the config a schema from where it can be easily validated?

dispatcher/brokers/pg_notify.py
```diff
@@ -13,87 +17,194 @@
...
+
+    def __init__(
+        self,
+        config: Optional[dict] = None,
+        channels: Union[tuple, list] = (),
+        default_publish_channel: Optional[str] = None,
+    ) -> None:
+        """
+        channels - listening channels for the service and used for control-and-reply
+        default_publish_channel - if not specified on task level or in the submission
```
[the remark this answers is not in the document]
AlanCoding: Valid call-out, because it's true that we can, and have, get by requiring that `@task(queue=...)` specify some queue. If if not, the queue is passed at time of submission. There is 1 test that expects a error to be raised if no queue name is provided at any of these levels.

What I didn't love was that `Control` for control-and-reply always necessary required that the queue be passed. This causes usage headaches, which caused me to go in all kinds of directions, like this default, and https://github.com/ansible/dispatcher/pull/76, which would allow us to treat them like other tasks. I was always sloppy in AWX implementation.

Also, it would be _nice_ if you put the listener channel in the same place as the submission channel. The _only_ way to do that is this "default_publish_channel", because it can go in the config. That ways it's self-obvious, you see what channel things get submitted to, and loop up a few lines and see that you're listening to it. Otherwise, you're relying on your code matching your config, and these aren't in the same place.

dispatcher.yml
AlanCoding: Current model is that `@task` can specify "options" for the task, but not triggers for the task. This feels defensible for the time-being, but a single slate of specifications has appeal. It's just that I would do that by moving everything into the config, as opposed to moving everything into the code.

AlanCoding:
> Should have the config a schema from where it can be easily validated?

No and yes.

**yes** we can validate the schema. This is not jsonschema, but validating the content, which is a python dictionary by the point we would care to do this. Demo:

```python
from dispatcher.config import setup, settings
setup(from_file='dispatcher.yml')
pkw = settings.service['pool_kwargs']

import inspect
from dispatcher.pool import WorkerPool
signature = inspect.signature(WorkerPool.__init__)
parameters = signature.parameters
valid_kwargs = [k for k, v in parameters.items() if v.kind != inspect.Parameter.VAR_POSITIONAL and v.kind != inspect.Parameter.VAR_KEYWORD]

>>> valid_kwargs
['self', 'worker_id', 'process']
>>> set(pkw.keys()) - set(valid_kwargs)
set()
```

This assures that all keys are valid. This doesn't do type validation, that would be a general problem that I would rather take to be solved, as opposed to solve myself.

I hope this answers the first part that **no** we don't want to write out a schema somewhere, because this general method defines the schema we want.

What we're lacking is an articulation of the value-add from doing fancy `inspect` stuff like this, as opposed to just letting the code go error. Possible answers:

1. We could use this inspection to construct a schema dynamically. Put it where then?
2. Validate the config schema in the publisher, not just in the service. But in that case, we could just initialize the objects and let the error happen that way, which would be simpler and more reliable.

dispatcher/brokers/base.py
AlanCoding: I think I got it now. I pushed a commit that makes this a protocol. It passes the linters (doesn't really affect the logic) and I think this is how it's supposed to be used.

Alex-Izquierdo:
> > Should have the config a schema from where it can be easily validated?
> 
> No and yes.
> 
> **yes** we can validate the schema. This is not jsonschema, but validating the content, which is a python dictionary by the point we would care to do this. Demo:
> 
> ```python
> from dispatcher.config import setup, settings
> setup(from_file='dispatcher.yml')
> pkw = settings.service['pool_kwargs']
> 
> import inspect
> from dispatcher.pool import WorkerPool
> signature = inspect.signature(WorkerPool.__init__)
> parameters = signature.parameters
> valid_kwargs = [k for k, v in parameters.items() if v.kind != inspect.Parameter.VAR_POSITIONAL and v.kind != inspect.Parameter.VAR_KEYWORD]
> 
> >>> valid_kwargs
> ['self', 'worker_id', 'process']
> >>> set(pkw.keys()) - set(valid_kwargs)
> set()
> ```
> 
> This assures that all keys are valid. This doesn't do type validation, that would be a general problem that I would rather take to be solved, as opposed to solve myself.
> 
> I hope this answers the first part that **no** we don't want to write out a schema somewhere, because this general method defines the schema we want.
> 
> What we're lacking is an articulation of the value-add from doing fancy `inspect` stuff like this, as opposed to just letting the code go error. Possible answers:
> 
> 1. We could use this inspection to construct a schema dynamically. Put it where then?
> 2. Validate the config schema in the publisher, not just in the service. But in that case, we could just initialize the objects and let the error happen that way, which would be simpler and more reliable.

I agree that using JSON Schema might be overkill. However, my main concern is not how the validation is done but rather ensuring that validation happens effectively.

The primary argument for validation is the "fail faster, fail earlier" principle, which improves both error handling and user experience. If an error in the configuration can be detected early, we can provide a clear and specific error message explaining what is wrong and why it is not accepted. This prevents unnecessary execution and debugging effort.

I also disagree with the idea of simply initializing the objects and letting the error occur naturally. This approach may lead to errors that are harder to debug or, worse, result in silent failures or unexpected behaviors.

Your approach is certainly creative, but I see some potential issues. It would not account for optional parameters, and it relies on a strict 1:1 relationship between the configuration and function/object signatures. This limits flexibility and, in my opinion, introduces a partial coupling of concerns.

For these reasons, I advocate for an independent configuration that can be validated as early as possible.

Alex-Izquierdo:
I would like to clarify that I'm not going to block the PR on this matter. In one way or another, it is something that can be addressed perfectly in later iterations. I'm just sharing my PoV on how it could be. :)

AlanCoding:
Auto-gen schema could help for documentation and versioning. We could put this in the repo and check for diff in checks. If there's a diff, then we can bump a version.

This is to help with a real problem, which is versioning. My intent was to cut a version (PyPI, or just a tag) before merging this because it requires changes to the eda-server branch. Auto-generating a spec file (that is enforced) would document when and how something changed to expected schema.

But failing earlier doesn't make sense to me. Initializing the objects from the factories (from settings) is "free", and can be done in any context.

AlanCoding:
This is moving a bit fast, but I went ahead and pushed a commit that matches the description of my last comment.

I appreciate that this would be better as jsonschema. But that requires a general tool to convert type hints to jsonschema, there's no point in us maintaining that logic. Let's file an issue for it. Current format is kind of human readable, but jsonschema would be better.

tests/integration/test_main.py
```diff
@@ -24,19 +26,27 @@ async def wait_to_receive(dispatcher, ct, timeout=5.0, interval=0.05):
...
 
     assert apg_dispatcher.pool.finished_count == 1
 
 
 @pytest.mark.asyncio
-async def test_run_decorated_function(apg_dispatcher, conn_config):
-    assert apg_dispatcher.pool.finished_count == 0
+async def test_run_decorated_function(apg_dispatcher, test_settings):
+    clearing_task = asyncio.create_task(apg_dispatcher.pool.events.work_cleared.wait())
+    test_methods.print_hello.apply_async(settings=test_settings)
```
Alex-Izquierdo, line 39: Just for reference: https://github.com/ansible/dispatcher/issues/83, not a blocker for merging. 

I believe you mentioned at some point that you don't want to run checkers on tests. I would agree that in some cases can be annoying or not useful, but in that cases we can just use exceptions/ignores
I think treating the test code with the same rules has some benefits: on one hand the test would allow us to consume the library as a client and detect issues that we could not detect in other way as well as enhance the development or debugging of tests.

tests/unit/conftest.py
Alex-Izquierdo, line 1: would not be better if we just remove the file? Not a blocker for merging, anyway.

tests/unit/test_config.py
```diff
@@ -1,3 +1,52 @@
...
+    settings = LazySettings()
+    with pytest.raises(Exception) as exc:
+        settings.brokers
+    assert 'Dispatcher not configured' in str(exc)
+
+
+def test_configured_settings():
+    settings = LazySettings()
+    settings._wrapped = DispatcherSettings({'version': 2, 'brokers': {'pg_notify': {'config': {}}}})
+    'pg_notify' in settings.brokers
```
Alex-Izquierdo, line 21: I think you forgot the assertion. 
```suggestion
    assert 'pg_notify' in settings.brokers
```

dispatcher/brokers/base.py
```diff
@@ -0,0 +1,13 @@
+from typing import Optional, Protocol
+
+
+class BaseBroker(Protocol):
+    async def aprocess_notify(self, connected_callback=None): ...
```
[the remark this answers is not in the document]
Alex-Izquierdo: ok, give it a quick try if you want, otherwise let's go ahead with it.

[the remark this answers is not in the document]
AlanCoding: Wow, it seems that https://github.com/python/mypy/pull/18422 has the answer here.

Because we are using a `Protocol` that doesn't define the method body, the method has no `yield` statement. There is special type treatment in cPython based on whether a method has a yield statement or not... obviously problematic for protocols!

So the suggested solution is to change `async def` --> `def`, but this appears to only be needed for the protocol class, as expected.

Hopefully I never have to deal with this mess again.

dispatcher/registry.py
```diff
@@ -79,23 +81,21 @@ def get_async_body(
 
         return body
 
-    def apply_async(self, args=None, kwargs=None, queue=None, uuid=None, connection=None, config=None, **kw) -> Tuple[dict, str]:
+    def apply_async(self, args=None, kwargs=None, queue=None, uuid=None, settings: LazySettings = global_settings, **kw) -> Tuple[dict, str]:
         queue = queue or self.submission_defaults.get('queue')
-        if not queue:
```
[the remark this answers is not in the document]
AlanCoding: There's no type hint because it's old code adopted here...

This was removed because additional handling was added to the `Broker` class, and specifically, because I wanted to add a default publishing queue on the broker. So the expectation changes from an exception being raised here to being raised later in `broker.publish_message(channel=queue, message=json.dumps(obj))` which would then call `get_publish_channel`. What was being enforced here (obviously that final value of `queue` is not `None`) is enforced there.

Actually, I will make this clear by type hints, that `get_publish_channel` optionally accepts a channel, but unconditionally returns a channel. That's the point of its existence - as a default enforcer.

tests/integration/test_main.py
AlanCoding: I agree we should run checkers on tests, I just don't believe they should be the same checkers as on the code. Tests need to pass objects of the wrong type... because they're testing what it does in that case.

https://github.com/ansible/dispatcher/issues/83 seems like a little bit of a different subject to me, and it is a real problem. Firstly, it seems we need to do the obvious of completing type hints for the methods here. After that, I'm unclear about what the best practice would be. There's a lot you need to do to carry over methods that you wrap, but `@task` doesn't technically wrap.

dispatcher/brokers/base.py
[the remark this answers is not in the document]
Alex-Izquierdo: I did a quick test but using pyright, which I feel handles better the types that mypy. I don't know how much is accurate or good practice to set as sync method something that is actually async... :(

Alex-Izquierdo (approved):

dispatcher/brokers/base.py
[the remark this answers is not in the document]
AlanCoding: This is dancing around the bleeding edge, and because of that we won't get a "right" answer. But I'll push a commit that changes the Protocol method so it remains async. That matches both our personal preferences better.

Alex-Izquierdo (approved):

PR #69, opened 2025-02-11
chore: rename "dispatcher" to "ansible-dispatcher"

Alex-Izquierdo:
Preparing the work for an eventual rename to ansible-dispatcher || ansible_dispatcher || Ansible Dispacher
Includes the rename of the package and all the references in scripts, docs and code. It also assumes the repository is renamed to the same name.


AlanCoding:
I was really hoping we could get other open patches merged in before the file movement changes. When you're ready to do this, I want to clear things out, make the changes, and get them in quickly.

Alex-Izquierdo:
np.

ansible_dispatcher/__init__.py
webknjaz, line 1: I recommend putting it under `src/` to avoid surprises when testing.

ansible_dispatcher/cli.py
```diff
@@ -6,7 +6,7 @@
 
 import yaml
 
-from dispatcher.main import DispatcherMain
+from ansible_dispatcher.main import DispatcherMain
```
webknjaz, line 9: Why not just use relative imports everywhere?

pyproject.toml
```diff
@@ -25,10 +32,10 @@ classifiers = [
 dependencies = ["pyyaml"]
 
 [project.scripts]
-dispatcher-standalone = "dispatcher.cli:standalone"
+dispatcher-standalone = "ansible_dispatcher.cli:standalone"
```
webknjaz, line 35: should the executable also be renamed?

Makefile
```diff
@@ -13,10 +13,10 @@ stop-postgres:
 clean:
 	find . -type f -regex ".*\.py[co]$$" -delete
 	find . -type d -name "__pycache__" -delete
-	rm -rf dispatcher.egg-info/
+	rm -rf ansible_dispatcher.egg-info/
```
webknjaz, line 16: Instead of tracking things manually, it's usually easier to just invoke `git clean`.

ansible_dispatcher/__init__.py
AlanCoding: neither ansible_base nor awx use a `src/` folder, which are deeply tied to this, so I would disfavor that proposal. Being consistently wrong is better than being inconsistently right.

ansible_dispatcher/cli.py
AlanCoding: That would be fine with me. As long as everything still worked.

pyproject.toml
AlanCoding: I would like to add another `dispatcherctl` executable eventually. This does seem like a long name.

Tangential: I'm also interested to see if the entrypoints can be put in an "extras" package. So if someone doesn't install the extra (they run the service from their own script), then they don't get the entrypoint.

AlanCoding:
I'm good for you to land this soon. I'd also like to handle https://github.com/ansible/dispatcher/issues/68

So let's coordinate, so that once you put the work in to finalize it, we can merge right away. Then I'd like to also efficiently handle issue 68 right after that.

ansible_dispatcher/cli.py
Alex-Izquierdo: I agree, but I would prefer to just rename the existing pattern in this PR and make the change in a different PR.

Alex-Izquierdo:
better to wait for a final decision about the name. It does not seem to be ansible-dispatcher.

PR #70, opened 2025-02-11
Separate process management out of pool

AlanCoding:
This is pre-work before introducing use of "forkserver".

As we introduce that new method of forking, my plan is to preserve the current logic. Instead of replacing, this will be a configurable class.

For now, this enabling change attempts to quarantine all interactions with the python `multiprocessing` library into a new `process` module. This leaves `pool` as a module highly specialized on asyncio task management and balancing of capacity.

### DIFF
--- a/dispatcher/process.py
+++ b/dispatcher/process.py
@@ -1,14 +1,14 @@
 import asyncio
 import multiprocessing
-from typing import Iterable, Optional, Union
+from typing import Callable, Iterable, Optional, Union
 
 from dispatcher.worker.task import work_loop
 
 
 class ProcessProxy:
-    def __init__(self, args: Iterable, finished_queue: multiprocessing.Queue) -> None:
+    def __init__(self, args: Iterable, finished_queue: multiprocessing.Queue, target: Callable = work_loop) -> None:
         self.message_queue: multiprocessing.Queue = multiprocessing.Queue()
-        self._process = multiprocessing.Process(target=work_loop, args=tuple(args) + (self.message_queue, finished_queue))
+        self._process = multiprocessing.Process(target=target, args=tuple(args) + (self.message_queue, finished_queue))
 
     def start(self) -> None:
         self._process.start()
@@ -46,8 +46,8 @@ class ProcessManager:
             self._loop = asyncio.get_event_loop()
         return self._loop
 
-    def create_process(self, args: Iterable) -> ProcessProxy:
-        return ProcessProxy(args, self.finished_queue)
+    def create_process(self, args: Iterable[int | str], **kwargs) -> ProcessProxy:
+        return ProcessProxy(args, self.finished_queue, **kwargs)
 
     async def read_finished(self) -> dict[str, Union[str, int]]:
         message = await self.get_event_loop().run_in_executor(None, self.finished_queue.get)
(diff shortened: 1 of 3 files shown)


AlanCoding:
Let me explain what the intent here is - yes, the purpose is so that we can later _swap out_ the `ProcessManager` class for different classes that implement these things differently. What other implementations would we use? You might ask...

1. As is well-established, the forkserver implementation described in https://github.com/ansible/dispatcher/issues/39, research patch at https://github.com/ansible/awx/compare/devel...AlanCoding:awx:forkserver, from the patch, you can see this mainly requires setting some `multiprocessing` global context settings
2. Other proposal, which still stands, is using janus https://github.com/ansible/dispatcher/issues/3, experimental diff at https://github.com/ansible/dispatcher/compare/main...AlanCoding:dispatcher:janus, and you can see from the patch that we're not changing anything about the `Process` object, only the `Queue` creation, and even then, in a way that's fairly trivial to swap out with this class structure. While this didn't work before, it might be due to the threading-related issues we were already warned about, and _combining_ this with (1) might make it "just work".

Both of these are very viable with this class structure. I think we should add them, and test them. One, because we would change the default and that solves an urgent problem. But even after that, it's like driving around with a spare tire in your trunk. My memory is that celery had major issues that happened once it started playing with novel forking methodologies.  Once this stabilizes, we should keep around some options, just in case.

Alex-Izquierdo (commented):
LGTM, have you done some manual testing?

dispatcher/process.py
```diff
@@ -0,0 +1,54 @@
+import asyncio
+import multiprocessing
+from typing import Iterable, Optional, Union
+
+from dispatcher.worker.task import work_loop
+
+
+class ProcessProxy:
+    def __init__(self, args: Iterable, finished_queue: multiprocessing.Queue) -> None:
```
webknjaz, line 9: The args type is incomplete. Would this be more accurate?
```suggestion
    def __init__(self, args: Iterable[int | str], finished_queue: multiprocessing.Queue) -> None:
```

AlanCoding: Looks good, included.

AlanCoding:
I added some simple tests. I don't see a whole lot of value (almost trivial), but I like how this is informative, how it steps-through the layers of code construction.

pb82 (approved):
LGTM. So `ProcessProxy` represents an individual process and `ProcessManager` is the interface for interacting with proesses (creating, reading results)

dispatcher/process.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def get_event_loop(self):
+        if not self._loop:
+            self._loop = asyncio.get_event_loop()
+        return self._loop
+
+    def create_process(self, args: Iterable[int | str], **kwargs) -> ProcessProxy:
+        return ProcessProxy(args, self.finished_queue, **kwargs)
+
+    async def read_finished(self) -> dict[str, Union[str, int]]:
```
pb82, line 52: is this returning a single process result? how do we arrive at the type `dict[str, Union[str, int]]`?

AlanCoding: This reads from the finished queue, asynchronously. Each message will be that `dict` type, meaning, that's the data structure expected that the worker put into the queue. So that'll be in the worker code, but here it comes from the .get.

Alex-Izquierdo (approved):

PR #75, opened 2025-02-13
Drop the unnecessary `wheel` entry from build deps

webknjaz:
It was never needed to begin with, as `setuptools` would auto-depend on it when building wheels in the past, anyway. But nowadays, `wheel` moved inside of `setuptools` even (during last year's PyCon US Sprints, and I was helping a bit, actually).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 Repository = "https://github.com/ansible/dispatcher"
 
 [build-system]
-requires = ["setuptools", "wheel"]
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):
cargo-culted stuff

Alex-Izquierdo (approved):

webknjaz:
Yeah, I think it was me who initially put `wheel` into the `setuptools`' docs examples. By the time I fixed the mistake enough people copied it...

PR #77, opened 2025-02-15
Worker auto-scaling

AlanCoding:
Fixes https://github.com/ansible/dispatcher/issues/50

Went through some revisions, had a couple of bugs to work through, like forgetting to multiple nanoseconds back to seconds. Worked through those, and organized into a coherent method structure.

You can run the demo script with this and see workers scale up. It actually doesn't throw enough to get up to the 12 max workers, but if you run it multiple times you can see the message that it hits max workers. Then after some-seconds after that, it scales down a ton of workers in a quick burst. I kind of wanted this to be spread out more, but I think it's correct given the profile of the workload.

### DIFF
--- a/dispatcher/service/pool.py
+++ b/dispatcher/service/pool.py
@@ -27,8 +27,16 @@ class PoolWorker:
         self.exit_msg_event = asyncio.Event()
 
     async def start(self) -> None:
+        if self.status != 'initialized':
+            logger.error(f'Worker {self.worker_id} status is not initialized, can not start, status={self.status}')
+            return
         self.status = 'spawned'
-        self.process.start()
+        try:
+            self.process.start()
+        except Exception:
+            logger.exception(f'Unexpected error starting worker {self.worker_id} subprocess, marking as error')
+            self.status = 'error'
+            return
         logger.debug(f'Worker {self.worker_id} pid={self.process.pid} subprocess has spawned')
         self.status = 'starting'  # Not ready until it sends callback message
 
--- a/tests/unit/service/test_pool.py
+++ b/tests/unit/service/test_pool.py
@@ -1,4 +1,6 @@
 import time
+import asyncio
+from unittest import mock
 
 import pytest
 
@@ -98,3 +100,20 @@ async def test_scale_down_condition(test_settings):
     # Same number of workers but one worker has been sent a stop signal
     assert len(pool.workers) == 3
     assert set([worker.status for worker in pool.workers.values()]) == {'ready', 'stopping'}
+
+
+@pytest.mark.asyncio
+async def test_error_while_scaling_up(test_settings):
+    """It is always possible that we fail to start workers due to OS errors. This should not error the whole program."""
+    pm = ProcessManager(settings=test_settings)
+    pool = WorkerPool(pm, min_workers=1, max_workers=1)
+
+    pool.queued_messages = [{'task': 'waiting.task'}]
+    for i in range(3):
+        await pool.scale_workers()
+    assert len(pool.workers) == 1
+
+    with mock.patch('dispatcher.service.process.ProcessProxy.start', side_effect=RuntimeError):
+        await pool.manage_new_workers(asyncio.Lock())
+
+    assert set([worker.status for worker in pool.workers.values()]) == {'error'}


dispatcher/pool.py
```diff
@@ -374,6 +441,10 @@ async def process_finished(self, worker, message) -> None:
         if 'timeout' in message:
             self.events.timeout_event.set()
 
+        if not self.get_unblocked_message():
+            # if nothing is in the queue, it may be clear to scale down
+            self.events.management_event.set()
```
AlanCoding, line 446: Had to write this to find out this is completely the wrong way to trigger scale-down. This would totally fail to scale down after getting a burst of work that goes quiet afterwards. In AWX, it scales down based on a periodic check. This should do the same thing. Just add a timeout to this:

```python
await self.events.management_event.wait(timeout=self.scale_down_check_frequency)
```

This trigger needs to be removed completely. It does no good.

dispatcher/service/pool.py
```diff
@@ -130,20 +147,62 @@ async def start_working(self, dispatcher) -> None:
...
+        ct = 0
+        for worker in self.workers.values():
+            if worker.current_task:
+                ct += 1
+        return ct
+
+    def should_scale_down(self):
+        worker_ct = len([worker for worker in self.workers.values() if worker.status == 'ready'])
+        if self.last_used_by_ct.get(worker_ct):
+            delta = (time.monotonic_ns() - self.last_used_by_ct[worker_ct]) / 1.0e-9
```
AlanCoding, line 160: > / 1.0e-9

Oh, well there's a problem right there.

AlanCoding:
I actually think this might be working now. I struggle thinking about how to formalize it better. I can/should write some basic tests, but it's going to be hard to capture the full mechanics.

art-tapin (approved):
I've reviewed your auto-scaling PR and I'm impressed with how comprehensive the implementation is! The separation of worker lifecycle stages, the time handling improvements, and the flexible configuration options are all great additions.

The code mostly looks ready to merge, but I found a few things worth addressing:

1. The docstring for `test_scale_down_condition` seems to be incorrectly copied from another test - it describes scaling up behavior when the test is actually verifying scale-down.

2. There are a couple of potentially important things to consider:
   - It looks like you're using the management lock for scale-down operations but not for scale-up. Could this lead to race conditions if multiple threads try to scale up at once?
   - I don't see validation for config parameters - do we need to ensure `min_workers` ≤ `max_workers`?
   - How does the system handle errors during worker startup/shutdown? Are those propagated back to the management loop?

3. For future improvements (not blockers):
   - Some methods iterate through all workers/messages - any performance concerns for large deployments?
   - Would it be helpful to add some guidance on how to tune the auto-scaling parameters for different workloads?

I ran the tests locally and they're passing, which is a good sign! The integration test for scaling up to full capacity is particularly nice to see.

LGTM :+1: 

:1st_place_medal:

AlanCoding:
> It looks like you're using the management lock for scale-down operations but not for scale-up. Could this lead to race conditions if multiple threads try to scale up at once?

My intended use of the management lock is only for cases that deal with placement of a task in a worker.  Scaling up adds a worker (which isn't available right away) and doesn't place any tasks.

Scaling down is different, because what if some code was trying to place a task around the same time we decide to scale down? That could lead to placing a task in an exiting worker. Both cases are "consuming" a worker, regarding capacity. Scaling up doesn't.

There's also no possibility of multiple threads scaling up at the same time, because this logic is running in a task that is effectively a singleton. If, later, we found performance reasons to parallelize this, we would use `asyncio.gather` for the forking or something that still doesn't introduce multiple concurrent tasks evaluating the scale-up condition. This is a different concern from the worker placement issue.

Good question.

AlanCoding:
> The docstring for test_scale_down_condition seems to be incorrectly copied from another test

Good find. I pushed a commit for this. There's a train of thought that the tests were following, and this was dropping the ball.

AlanCoding:
> How does the system handle errors during worker startup/shutdown? Are those propagated back to the management loop?

Also good point. Specifically looking at startup here. We could get an OOM error originating from the OS. This was not considered.

I'll push a commit. This library is still very light on exception testing, but this is a good specific case I should go ahead and add it. I'll set the worker status to "error" and then it'll continue about enforcing the existing policies. These policies will still ensure the pid is not running. It will slow things down, but I would want things to be slowed down when in the vicinity of OOM scenarios.

AlanCoding:
> Some methods iterate through all workers/messages - any performance concerns for large deployments?

Since these are simple python loops, I think we are far, far, away from that point. However, I'm not very concerned with scaling logic because that's almost always `O(N)` regarding the worker count.

What is more concerning is this logic for basic task placement:

https://github.com/ansible/dispatcher/blob/5c479218938e785a313b71f9afc8dc2561e6ee0e/dispatcher/service/pool.py#L312-L317

This is because:
 - running a task should be vastly more frequent than scaling up a new worker
 - this particular code has a worst-case of something like `O(N^2)` with the worker count because `message_is_blocked` requires peeking at all running tasks

I have been thinking about that case, and related to some refactors like https://github.com/ansible/dispatcher/issues/85, we should change the loop so that it gets a set of hashes to compare against. And then that will take care of it. Worker scaling logic shouldn't need any adjustments, but, managing the triggers for the task is much more prone to issues.

PR #84, opened 2025-02-20
Fix integration bugs between scheduled producer, `@task` decorator and registry

AlanCoding:
Fixes https://github.com/ansible/dispatcher/issues/82

There's 3 semi-related things going on, so I'll go through in the order in which I approached them.

1. Scheduled tasks didn't properly respect `bind=True`, because all the registry logic was never adopted in the ScheduledProducer, like, at all. But the fix _for this_ assuming that imports are already in place, is to move the application of the "bind" argument, specifically to the worker. Why only this? It just happens to be the only option that affects behavior on the worker, and does not affect behavior in the parent process. It's just... the only one that does that.
2. Remember that qualifier "assuming that imports are already in place". Well this isn't a good assumption. So if they're not, we'll have a "miss", not finding the method in the registry even through it has a `@task()` on it. The solution is to double-check an instance variable on the registry to see if the import of the method affects the registry itself. The test is super duper cheeky. But I find these methods acceptable, because it's the only way to get the coverage we need, in an isolated way. And the tests are "our" space for improving software quality. So if I need to add a special purpose module, I don't see why I can't.
3. Remember how the "bind" option is the only one applied in the worker. What about all the other options, like `on_duplicate`? Doesn't `ScheduledProducer` need to apply those too? Unfortunately I don't have a solution, because this leaves me at a philosophical crisis. More-and-more, we have leaned into a strict environment separation between the dispatcher and the Django app. We don't set up Django in the dispatcher. The user can, that's not our business. But the dispatcher _shouldn't_ and maybe _can't_ import the scheduled methods in the main asyncio code. It should only know them by their name. Forcing importing will bring in dirty stuff. So for now I'm adding a **workaround** so that you can put method options directly in the schedule spec.

### DIFF
--- a/dispatcher/producers/scheduled.py
+++ b/dispatcher/producers/scheduled.py
@@ -14,8 +14,9 @@ class ScheduledProducer(BaseProducer):
 
     async def start_producing(self, dispatcher) -> None:
         for task_name, options in self.task_schedule.items():
-            per_seconds = options['schedule']
-            schedule_task = asyncio.create_task(self.run_schedule_forever(task_name, per_seconds, dispatcher))
+            submission_options = options.copy()
+            per_seconds = submission_options.pop('schedule')
+            schedule_task = asyncio.create_task(self.run_schedule_forever(task_name, per_seconds, dispatcher, submission_options))
             self.scheduled_tasks.append(schedule_task)
             schedule_task.add_done_callback(dispatcher.fatal_error_callback)
         if self.events:
@@ -24,13 +25,16 @@ class ScheduledProducer(BaseProducer):
     def all_tasks(self) -> list[asyncio.Task]:
         return self.scheduled_tasks
 
-    async def run_schedule_forever(self, task_name: str, per_seconds, dispatcher) -> None:
+    async def run_schedule_forever(self, task_name: str, per_seconds, dispatcher, submission_options) -> None:
         logger.info(f"Starting task runner for {task_name} with interval {per_seconds} seconds")
         while True:
             await asyncio.sleep(per_seconds)
             logger.debug(f"Produced scheduled task: {task_name}")
             self.produced_count += 1
-            await dispatcher.process_message({'task': task_name, 'uuid': f'sch-{self.produced_count}'})
+            message = submission_options.copy()
+            message['task'] = task_name
+            message['uuid'] = f'sch-{self.produced_count}'
+            await dispatcher.process_message(message)
 
     async def shutdown(self) -> None:
         logger.info('Stopping scheduled tasks')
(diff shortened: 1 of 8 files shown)


dispatcher/producers/scheduled.py
```diff
@@ -24,13 +25,16 @@ async def start_producing(self, dispatcher) -> None:
...
+    async def run_schedule_forever(self, task_name: str, per_seconds, dispatcher, submission_options) -> None:
         logger.info(f"Starting task runner for {task_name} with interval {per_seconds} seconds")
         while True:
             await asyncio.sleep(per_seconds)
             logger.debug(f"Produced scheduled task: {task_name}")
             self.produced_count += 1
-            await dispatcher.process_message({'task': task_name, 'uuid': f'sch-{self.produced_count}'})
+            message = submission_options.copy()
+            message['task'] = task_name
+            message['uuid'] = f'sch-{self.produced_count}'
```
Alex-Izquierdo, line 36: Is not weird to use `uuid` for something that is not actually an uuid? 
https://www.rfc-editor.org/rfc/rfc9562.html
Would you consider a new issue to eventually redesign the schema of the message?

tests/unit/service/producers/test_scheduled_producer.py
```diff
@@ -0,0 +1,42 @@
...
+def test_scheduled_producer_with_options():
+    producer = ScheduledProducer({
+        'tests.data.methods.print_hello': {
+            'schedule': 0.1,
+            'on_duplicate': 'queue_one'
+        }
+    })
+
+    loop = asyncio.get_event_loop()
+    try:
```
Alex-Izquierdo, line 39: Should not we use here `with pytest.raises(ItWorked)` ?

dispatcher/producers/scheduled.py
AlanCoding: > Would you consider a new issue to eventually redesign the schema of the message?

Yes. At this point in code we can introduce new fields without risking the 8,000 char limit. The presumed direction this is going in is that we would add additional fields to identify the origination of the message to replace what we were getting from `uuid`, which is log quality.

Some people might disagree, but I would rather keep `uuid` out of logs if it is a "real" UUID. One case for https://github.com/ansible/dispatcher/issues/8 is to pass logging identifiers that are bound from the original request or job. This requires a custom formatter on the app side. For out-of-the-box dispatcher, there's no reason to believe that a uuid is any more helpful for log tracing than origin information.

tests/unit/service/producers/test_scheduled_producer.py
AlanCoding: :+1:  looks good, just had to check that it works. Will push next commit.

Alex-Izquierdo (approved):

PR #88, opened 2025-02-23
Send reply messages from the main process

AlanCoding:
This changes the implementation of the control-and-reply pattern so that reply messages are sent directly from the main asyncio code, instead of sending them back using the workers in synchronous code.

I had previously attempted this and failed, discovering that it would hang when trying to send the reply. After giving it another shot today, I discovered that this behavior is described in https://github.com/psycopg/psycopg/issues/340 upstream. This implements a solution that is compliant with the suggested use of the psycopg library, which is that we exit the `async for` loop before sending the reply, and then continue the `while True:` loop which already exists.

### Why did I not do this before? (other than the obvious reason it didn't work)

There are some trade-offs involved, in the AWX dispatcher, we used a different connection to send the reply. This is a bad idea, but more-so for error handling and consistency reasons, as opposed to timing reasons. Action on that separate connection _generally_ caused blocking problems, but the act of sending a reply is not really the main culprit. I still had a thought that it would be better to farm this work to the workers.

### Why do those reasons not apply here?

As much as the reply message would have been blocking in the AWX dispatcher, this implementation would be _less_ blocking because it is bona fide asyncio code. If this does pose any delays in getting new messages, that's probably because we're sunk capacity-wise or with other errors. In those cases, getting debugging information (which the replies do) would be a higher priority than whatever those new messages say.

Also, I've been shifting further and further to the idea of _relying_ on the control-and-reply pattern to replace heartbeat logic. With that as a goal, this is undoubtedly the correct thing to do.

### Usability implications

With this, the demo script now runs in <1 second, and is rock-solid reliable. Before this, it didn't. That's because the demo over-and-over-again submitted more work than the 4 workers had capacity to handle (intentionally). Now, no "work" blocks the demo script logic. This is also relevant for eda-server as it uses the "alive" message as a replacement for some rq-worker stuff.

### DIFF
--- a/dispatcher/brokers/pg_notify.py
+++ b/dispatcher/brokers/pg_notify.py
@@ -70,6 +70,12 @@ class Broker:
         self.channels = channels
         self.default_publish_channel = default_publish_channel
 
+        # If we are in the notification loop (receiving messages),
+        # then we have to break out before sending messages
+        # These variables track things so that we can exit, send, and re-enter
+        self.notify_loop_active: bool = False
+        self.notify_queue: list = []
+
     def get_publish_channel(self, channel: Optional[str] = None) -> str:
         "Handle default for the publishing channel for calls to publish_message, shared sync and async"
         if channel is not None:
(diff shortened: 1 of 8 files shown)


dispatcher/brokers/pg_notify.py
```diff
@@ -18,11 +18,17 @@
...
+    if not connection.autocommit:
+        await connection.set_autocommit(True)
+    return connection
 
 
 def create_connection(**config) -> psycopg.Connection:
-    return psycopg.Connection.connect(**config)
+    connection = psycopg.Connection.connect(**config)
+    if not connection.autocommit:
+        connection.set_autocommit(True)
```
AlanCoding, line 30: For future readers, this _has_ a hazard, which is that sometimes you _want_ to send messages as a part of a transaction, for transactional integrity. This is documented - messages will not send if transactions are rolled back, and they will delay until the commit. This feature is actually used in AWX.

But in any case someone relies on that behavior, they need to use the connection factory. In other words, provide their own connection. This is only called if dispatcher, itself, has to create a connection for some reason, and is bypassed by passing a connection factory. Specifically, this is how we would use the Django connection without creating a new connection. _That_ connection may keep autocommit off.

dispatcher/brokers/pg_notify.py
```diff
@@ -18,14 +18,21 @@
...
+    return connection
 
 
 class Broker:
+    NOTIFY_SYNTAX = 'SELECT pg_notify(%s, %s);'
```
Alex-Izquierdo, line 35: Can I suggest a more self descriptive name like "PG_NOTIFY_SQL_TEMPLATE" or "PG_NOTIFY_QUERY_TEMPLATE" ?

dispatcher/control.py
```diff
@@ -35,6 +35,7 @@ async def process_message(self, payload, producer=None, channel=None):
         self.received_replies.append(payload)
         if self.expected_replies and (len(self.received_replies) >= self.expected_replies):
             self.events.exit_event.set()
+        return (None, None)
```
Alex-Izquierdo, line 38: The type checking of the function should have the return value.

Alex-Izquierdo (approved):

PR #89, opened 2025-02-24
Add method for synchronous listening

AlanCoding:
Kind of related to https://github.com/ansible/dispatcher/issues/46 but I really thought I had an issue on this particular thing. Maybe not.

I'm not happy with how messy this code is, but that's why I wasn't happy with the AWX dispatcher in the first place. I really tried other solutions, and was not happen with any of them. The `add_notify_handler` doesn't work (waits for query to finish before notifying). I really wanted to copy the methods from the pulpcore solution:

https://github.com/pulp/pulpcore/blob/1dbb7d46bd18ca8934f56ebcab8b59b550ed7121/pulpcore/tasking/worker.py

But that introduces _more_ complexity than what this does, and operates at a lower-level, which I wish to avoid.

This works and solves a problem, so I should add it. Coupled with https://github.com/ansible/dispatcher/pull/88, that will get the process and connection management in the final intended form for the full control-and-reply cycle.


dispatcher/brokers/pg_notify.py
```diff
@@ -25,6 +26,29 @@ def create_connection(**config) -> psycopg.Connection:
...
+            ns = conn.wait(psycopg.generators.notifies(conn.pgconn))
+        except psycopg.errors._NO_TRACEBACK as ex:
+            raise ex.with_traceback(None)
+    for pgn in ns:
+        if hasattr(conn.pgconn, '_encoding'):
+            # later, like 3.2+ versions
+            enc = conn.pgconn._encoding
+        else:
+            # earlier versions
+            enc = psycopg._encodings.pgconn_encoding(conn.pgconn)
```
AlanCoding, line 47: This generalizes relatively recent work from https://github.com/ansible/awx/pull/15705

AlanCoding:
I take it back, this should be done better.

AlanCoding:
Pushed, uses https://github.com/psycopg/psycopg/commit/f88dcbc0a314abca080715a840b508f1e08bc77d

AlanCoding:
This now requires a fairly recent version of psycopg3. But accepting that, we get something that is 5 gagillion times better than what we had before.

Alex-Izquierdo (approved):

PR #91, opened 2025-02-25
Start pattern to disallow untyped functions, MyPY updates

AlanCoding:
Coming from feedback in https://github.com/ansible/dispatcher/pull/88 and others, this starts to make sense.

I don't think this is a good idea for every library, and it may not be a good idea for all modules within the dispatcher. But the dispatcher is specialized as a code contract for other apps. The `dispatcher/control.py` module is a good example, as it originally came from AWX where we did none of this, but it is externally-facing, and this enables some meaningful progress towards integration in other apps. Speaking more specifically, there is value in the communication that `.control_with_reply` returns a list of dictionaries. This should be elaborated further to clarify that each dictionary in the list is data returned from a dispatcher service, since multiple running dispatcher services may reply.

Option `--ignore-missing-imports` was probably no longer needed after removing Django imports. This was possible after we clearly defined the swim lanes for dispatcher versus DAB.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
       - run: pip install mypy
       - run: pip install -e .[pg_notify]
       - run: python3 -m pip install types-PyYAML
-      - run: mypy --ignore-missing-imports dispatcher
+      - run: mypy dispatcher
 
   flake8:
     name: Run flake8
--- a/Makefile
+++ b/Makefile
@@ -19,4 +19,4 @@ linters:
 	black dispatcher/
 	isort dispatcher/
 	flake8 dispatcher/
-	mypy --ignore-missing-imports dispatcher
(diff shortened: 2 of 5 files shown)


AlanCoding:
I would still like to get this in. The other PR https://github.com/ansible/dispatcherd/pull/107 is also in the same code-quality category, and probably more important. But I can't think of a reason against either of them, and getting them in is low-risk and will help move forward.

AlanCoding:
Note that I filed https://github.com/ansible/dispatcherd/issues/109

So if you're hesitant like "this module still doesn't feel quite right", I agree with you. But with the open patches there's already a dog-pile happening on this code. It's got to be done step-by-step to a degree, or else you wind up throwing away code. And I know that's the case for issue 109 unless I can clear out other stuff.

pb82 (approved):
Should be good, CI is passing :+1:

PR #94, opened 2025-02-27
Pass settings in process, not pool, class. Use kwargs for `work_loop`

AlanCoding:
Totally random code organization nit this is trying to get out of the way -

When you look at it, it's clear that `WorkerPool` has no right to have `settings` in its kwargs. It does nothing with it other than to pass it on to the `ProcessManager` class. And the factories could just as well pass the settings to that class. So this makes that change.

The hard, ugly, part about this is that the arguments to `work_loop` get reorganized. Right now it's just an ordered list. I would be open to making these kwargs, I think that still works. The approach I'm going with now, is that we just order them in the order in which we build them. So that starts with data from `WorkerPool` and then makes its way in, from left-to-right, outer-to-inner.

### DIFF
--- a/dispatcher/factories.py
+++ b/dispatcher/factories.py
@@ -24,12 +24,11 @@ which is to avoid import dependencies.
 def process_manager_from_settings(settings: LazySettings = global_settings):
     cls_name = settings.service.get('process_manager_cls', 'ForkServerManager')
     process_manager_cls = getattr(process, cls_name)
-    return process_manager_cls()
+    return process_manager_cls(settings=settings)
 
 
 def pool_from_settings(settings: LazySettings = global_settings):
     kwargs = settings.service.get('pool_kwargs', {}).copy()
-    kwargs['settings'] = settings
     kwargs['process_manager'] = process_manager_from_settings(settings=settings)
     return WorkerPool(**kwargs)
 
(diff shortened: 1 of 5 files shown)


AlanCoding:
Using ordered arguments just got insane.

I changed that so we pass `kwargs` to match the `work_loop` method signature. This seems to work fine, we just have to consistently name the arguments in all cases. Fine.

dispatcher/service/pool.py
```diff
@@ -110,10 +108,9 @@ def __init__(self) -> None:
 
 
 class WorkerPool:
-    def __init__(self, max_workers: int, process_manager: ProcessManager, settings: LazySettings = global_settings):
+    def __init__(self, max_workers: int, process_manager: ProcessManager):
```
AlanCoding, line 111: This is going to conflict hard with the autoscaling PR.

But this is nice, because this bumps us down 1 kwarg, and that adds a ton of scalar kwargs. `WorkerPool.__init__` is one of the most modified method signature around, lol.

Alex-Izquierdo (approved):

dispatcher/factories.py
```diff
@@ -24,12 +24,11 @@
 def process_manager_from_settings(settings: LazySettings = global_settings):
     cls_name = settings.service.get('process_manager_cls', 'ForkServerManager')
     process_manager_cls = getattr(process, cls_name)
-    return process_manager_cls()
+    return process_manager_cls(settings=settings)
```
AlanCoding, line 27: This will conflict with another fix, but these are both additive changes, on this line. So add both things.

PR #99, opened 2025-02-27
Fix failure to pass process_manager_kwargs in factories

AlanCoding:
Also found in early AWX integration.

I added this to the settings but it didn't work, but then looking at the factories, I don't expect it was getting passed at all. So this is expected to fix that.


art-tapin (approved):
I've tested this and can confirm it fixes the issue by testing with custom preload_modules, which are now correctly loaded in worker processes after applying the fix. Tests coverage is :+1: 

LGTM! This is ready to merge. :1st_place_medal:

pb82 (approved):

PR #107, opened 2025-03-05
Add protocols for classes, prevent postgres-less test hang

AlanCoding:
Fixes https://github.com/ansible/dispatcher/issues/56

I wanted to combine these two things:
 - make tests finish in finite time if postgres is not running
 - define protocols for the top-level classes

The reason is that the first relates to the done callback in asyncio, so doing it refactors that interface, and makes the interfaces more coherent overall.


AlanCoding:
With last commit, I am able to get tests to finish (failing, erroring) without postgres running. So I'm going to add that to the scope of this, and I'm happy with things.

```
27 failed, 35 passed, 29 warnings, 28 errors in 14.11s
```

dispatcher/service/main.py
```diff
@@ -207,6 +199,11 @@ async def start_working(self) -> None:
                     logger.exception(f'Producer {producer} failed to start')
                     self.events.exit_event.set()
 
+                # TODO: recycle producer instead of raising up error
+                # https://github.com/ansible/dispatcherd/issues/2
+                for task in producer.all_tasks():
+                    ensure_fatal(task)
```
AlanCoding, line 205: @art-tapin I am trying to set something up here, and I want to call your attention to a specific decision.

For `WorkerPool`, I do the obvious thing, I call `ensure_fatal` right after a task is created. Why wouldn't you?

But for the brokers, we don't want to crash the program if a connection-related error happens. That broker self-check issue, we need to do retries and manage the broker. What will manage the broker? Probably the `DispatcherMain` with some periodic fallbacks in the main loop supporting this. Breaking it down, we have 2 scenarios - the broker loudly fails with a connection error, and a well known AWX bug - the client library (psycopg) loses the connection and gets no notification that this happened. The latter case is what we need periodic checks for. But for the former, we get an instant callback. Even in that case, we need to recycle the broker, so both cases need to feed into the same logic. It doesn't really matter which path you took to get there, you know what you need to do - recycle the broker and check the restart policy.

pb82:
@AlanCoding got a similar result when running the tests without postgres:

```
27 failed, 39 passed, 29 warnings, 28 errors in 14.45s
```

pb82 (approved):
With postgres running all tests are passing. Only thing I noticed was that some classes don't inherit from their protocol (e.g. WorkerPool). But from what I see, you mostly want to use them as function return types, so that's ok.

AlanCoding:
Maybe there are other feelings on this general practice, but protocols make the most sense to me if they are _not_ the base class. I was already aware from other projects that type-hinting has a massive problem with circular imports. Protocols as their own entity resolve this entirely. If `WorkerPool` fails to accurately implement the `WorkerPool` protocol, then it should be the case that MyPI gives errors.

Issue #96, opened 2025-02-27
Enforce UUID format, add "origin" information in message data


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

Broken off from comment from @Alex-Izquierdo 

https://github.com/ansible/dispatcher/pull/84#discussion_r1963607240

We have a `uuid` field, and you can see it being abused in the demo.

Instead, the demo should use some sort of "tag" for logging. And then producers should all set this to identify the source of the message, this is why I call it "origin" information.

To make the objectives concrete:
 - Nothing internal should ever set a "uuid" that is not `uuid4` format, right now it does, like `sch-4`, for the 4th message produced by the scheduled producer
 - Publishers are allowed to set a "uuid", but if it is not the expected format, issue a warning log and then replace it with a self-generated uuid

### Steps to reproduce

N/A

### Sugested feature result

Above

PR #108, opened 2025-03-05
[WIP] Standardize periodic task runners

AlanCoding:
Fixes https://github.com/ansible/dispatcher/issues/57

TODO: still need to apply this same tool to the `ScheduledProducer` because it's the 3rd version of the shared problem.

Before:
 - timeout watcher used 1 persistent task to watch for timeouts
 - delayed tasks used 1 task each

With this:
 - In all cases, either 0 or 1 task is running. That is, if there's no delayed task or task with a timeout, then we do not run a task. In all cases we rely on next-wakeup type logic, taken from the existing task timeout logic.

### DIFF
--- a/dispatcher/service/next_wakeup_runner.py
+++ b/dispatcher/service/next_wakeup_runner.py
@@ -30,7 +30,13 @@ def ensure_fatal(task: asyncio.Task) -> None:
 
 
 class NextWakeupRunner:
-    def __init__(self, wakeup_objects: Iterable[HasWakeup], callback: Callable[[], Coroutine[Any, Any, None]], name: Optional[str] = None):
+    """Implements a general contract to wakeup for next timestamp of a set of objects
+
+    For example, you have a set of schedules, each with a given period.
+    You want to run each on their period - this does that using one lazy asyncio task.
+    This is a repeated pattern in the code base with task schedules, delays, and timeouts
+    """
+    def __init__(self, wakeup_objects: Iterable[HasWakeup], callback: Callable[[], Coroutine[Any, Any, None]], name: Optional[str] = None) -> None:
         self.wakeup_objects = wakeup_objects
         self.callback = callback
         self.asyncio_task: Optional[asyncio.Task] = None
@@ -54,7 +60,7 @@ class NextWakeupRunner:
                 next_wakeup = nr_item
         return next_wakeup
 
-    async def background_task(self):
+    async def background_task(self) -> None:
         while not self.shutting_down:
             next_wakeup = self.get_next_wakeup()
             if next_wakeup is None:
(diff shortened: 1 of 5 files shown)


AlanCoding:
Fixed testing issues, now only remaining thing is to integrate with the scheduled producer.

dispatcher/producers/scheduled.py
```diff
@@ -1,49 +1,65 @@
...
+        self.dispatcher: Optional[DispatcherMain] = None
+        self.schedule_runner = NextWakeupRunner(self.schedule_entries, self.trigger_schedule)
         super().__init__()
 
-    async def start_producing(self, dispatcher) -> None:
+    async def trigger_schedule(self, entry: ScheduleEntry) -> None:
+        entry.mark_run()
+        self.produced_count += 1
+        message = entry.body.copy()
+        message['uuid'] = f'sch-{self.produced_count}'
```
pb82, line 38: I'm missing context of what this is used for, but is a predictable string here good enough as a uuid?

AlanCoding: This isn't the first time someone pointed this out. I filed https://github.com/ansible/dispatcherd/issues/96 to fix that.

dispatcher/service/next_wakeup_runner.py
```diff
@@ -0,0 +1,123 @@
...
+        """Initiates the asyncio task to wake up at the next run time
+
+        This needs to be called if objects in wakeup_objects are changed, for example
+        """
+        if self.process_wakeups(current_time=time.monotonic(), do_processing=False) is None:
+            # Optimization here, if there is no next time, do not bother managing tasks
+            return
+        if self.asyncio_task:
+            if self.asyncio_task.done():
+                self.mk_new_task()
```
pb82, line 110: is this to allow restarting the background task after it has been cancelled?

pb82 (approved):
Ok, I believe I understand the change: we now have all the logic for running delayed or periodic tasks in one abstraction:

`HasWakeup` marks a task as periodic (or delayed, which is the same but only one iteration). `NextWakeupRunner` processes periodic tasks and does not start a background process initially - it creates the 'supervising' task only when at least one periodic task has been created.

We're also using this new logic now in the worker pool.

dispatcher/service/next_wakeup_runner.py
AlanCoding: Not canceling, from the `return`. When the task does a `return` then it still exists as task. `.done()` will return `True`, and `.result()` will give the return value (which we don't care about). If there are no future wakeups, then the task does a `return`, so it's still hanging out but it's left. This is done specifically thinking about "delay" and "timeout" features which will only be used for a specific task once-in-a-while, and for some apps, never used at all. So I'd rather not have an active task in those cases.

Issue #85, opened 2025-02-20
Split out new classes - delayer, blocker, and queuer


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

The largest body of code we have is in `pool.py`, which is self-evident knowing the features it supports.

When the workers run out of capacity or a task is blocked, those go into `WorkerPool.queued_messages`. I want to re-think that statement.

Putting both blocked tasks and tasks waiting (due to lack of capacity) into the same line means that a whole new construct had to be introduced of `get_unblocked_message`. Say we put these into 2 different data structures:
 - `blocked_messages`
 - `queued_messages`

Then when capacity frees up, you are free to go from the first in the queued messages with no further questions asked. If there are none there, then you would then look in the blocked messages to see if anything has freed up.

In addition to that, the `main.py` has logic to support task delay. These are not held in a queue, but tasks are created for wake-up events. I have since been convinced that this would be better rewritten to create at most 1 task to support delayed tasks, see https://github.com/ansible/dispatcher/issues/57

This would mean that delayed tasks will be maintained in a data structure, maybe even a dequeue. So we'll call that:
 - `delayed_messages`

Now we have 3 clearly defined data structures which are all linear moving tasks through a state diagram. Hopefully, most tasks will skip all 3 of these. All the more reason... we should separate the code for handling these from the main line of main/pool code. Because of all that, this proposes to introduce 3 new types of classes:
 - `DelayedManager`
 - `BlockedManager`
 - `QueuedManager`

All 3 of these will have an internal store of tasks. The delayed manager will have a task associated, but neither of the others will.

We should have a protocol for all of these to define the interface. For this to work well with the `BlockedManager`, it will have to call out to the combined entity of the `WorkerPool` and the `QueuedManager` to get some generalized list of tasks which are "in-progress".

Because `DelayedManager` is a _producer_-like object, it should stay with the `main.py` code. The `BlockedManager` is the only one that is in-between, and could live with the main code or the pool code.

### Steps to reproduce

N/a

### Sugested feature result

n/a

AlanCoding:
Made this:

```mermaid
stateDiagram-v2
[...]

Issue #63, opened 2025-02-10
Implement a simple type of task priority


AlanCoding:
This will be unblocked once the auto-scaling is done in https://github.com/ansible/dispatcher/issues/50

That will result in a notion of "queue pressure" being established. When there is queue pressure, the code will enforce that a new worker is scaling up.

Priority will modify both:

 - when a task can be considered to contribute to queue pressure
 - what workers can be considered for task placement

The goal of the 2nd point would be to reserve extra workers that never process low priority tasks. An expression of the policy for a low-priority task might be:

> Task can not take the last remaining worker

> If task is queued, this does not constitute queue pressure (this task being queued does not cause scale-up events)

Current behavior is more-or-less, all tasks are highest priority.

AlanCoding:
I was just writing documentation with @Alex-Izquierdo as the `bind` argument, and I am realizing how cleanly this will form a "group" of task options for capacity management.

 - `on_duplicate` - have now, allows task shedding
 - `trigger_scale_up` - True / False, this task in the queue will trigger scaling up of the worker pool. This allows for low-prio tasks to be "patient", sit around and wait for capacity as opposed to growing the pool. Less clear: as a corollary to this, these may also be the last selected from the queue to run, which further minimizes scale-ups.
 - `reserve_workers` - int from 0 to `max_workers`. A value of `1` means that this tasks is not allowed to take the last available worker. Likewise, `2` means that it can only run if there are 3 free workers, leaving 2 reserve workers. And so on.

There is an interesting interaction between these, and we would need to write some "example" stories. For example, you could have a task that refuses to take the last worker, but still causes scale-up. So you always run the task (the task is important and time-sensitive), but it won't render the dispatcher temporarily unresponsive.

Internally, we need workers to send replies for control-and-reply communication. The reply is sending a single message and should be fast, but is _extremely_ time-sensitive. So this "built-in" task would leave no reserve workers, and likely be set to trigger scale-up.

AlanCoding:
Bouncing back from #85 I am trying to think how to best handle this new idea of "discard if there is no instantaneous capacity". Let me say the obvious things and then pick up after I've said them:

1. Add a new option `discard_for_capacity`
2. Seeing an overlap, change an existing option `trigger_scale_up` to no longer be boolean, but multi-choice

[...]

PR #111, opened 2025-03-06
Split logic for task blocking and queuing into a new module

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/85

This handles 2 of the classes mentioned there, becoming "processors" which is a new basic type of thing, here.

The 3rd type of the delayer will have its work... well... delayed, because https://github.com/ansible/dispatcherd/pull/108 will be heavily conflicting. That will become both a "processor" and a "next-wakup-runner-thing".

As such, I'm happy with this in terms of the code now. Yes, we could do _better_, and perhaps this was put together a little hastily at first. But this is an improvement and there is an outline of where we're going with it.

There are some performance aspects that could be better, but again, I don't think worse than before. I wanted to address that, but it needs a new class to replace passing around the `message` dictionary, and that's just too much right now.

----

But big picture your question is probably why we would be doing any of this? Because of https://github.com/ansible/dispatcherd/issues/63

This, along with auto-scaling, are the 2 pieces of groundwork to allow adding these capacity-management options to tasks. Yes, I already went ahead and added some new task options, but those are related to _blocking_ which can result in discarding, which is something to help with capacity management. Specifically the ability to have some tasks (but not others) avoid consuming a count of "reserve" workers is very important. This will allow the service to _virtually always_ maintain instantaneous worker capacity for more important work.

### DIFF
--- a/dispatcher/control.py
+++ b/dispatcher/control.py
@@ -80,7 +80,7 @@ class Control(object):
         await producer.start_producing(control_callbacks)
         for task in producer.all_tasks():
             # Make sure we catch errors
-            ensure_fatal(task)
+            ensure_fatal(task, exit_event=control_callbacks.events.exit_event)
 
         await producer.events.ready_event.wait()
 
--- a/dispatcher/producers/on_start.py
+++ b/dispatcher/producers/on_start.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 from typing import Union
 
+from ..protocols import DispatcherMain
 from .base import BaseProducer
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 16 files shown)


AlanCoding:
This is a bit of a mess now, sorry. The reason is that https://github.com/ansible/dispatcherd/pull/107 fixed that case (postgres not running) for the tests, but then broke it for running manually :(

The last pushes contain my fix for that, because the `.result()` doesn't work as I thought it had been advertised - to error the main thread. So contractually, this I think it cleaner as it passes asyncio objects and not entire dispatcherd objects.

pb82 (approved):

PR #112, opened 2025-03-06
Use SIGUSR1 for task cancellation (w/ tests)

art-tapin:
JIRA: [AAP-41128](https://issues.redhat.com/browse/AAP-41128)

## Description

Changed the dispatcher’s per-task cancellation signal from SIGTERM to SIGUSR1, preventing collisions with user/library SIGTERM handlers. Added an integration test that checks a user-defined SIGTERM handler is never triggered during normal cancellation. Added related docs to the design_notes.md.

## Testing

**Prerequisites**
- Run the dispatcher in a standalone mode (`make postgres`).

**Steps to test**
1. Run: `python -m pytest tests/integration/test_signal_handling.py`
2. Verify the new `test_sigusr1_cancel_avoids_sigterm` passes for both `fork` and `forkserver`, with no sign of `"SIGTERM was intercepted"` in logs.

**Notes**
Lambda-based tasks initially worked with SIGUSR1 but still encountered a SIGTERM at teardown. By defining a top-level `@task` and letting it exit quickly on cancellation, the test remains stable and confirms SIGUSR1 is used properly.

### DIFF
--- a/dispatcher/service/pool.py
+++ b/dispatcher/service/pool.py
@@ -1,5 +1,7 @@
 import asyncio
 import logging
+import os
+import signal
 import time
 from asyncio import Task
 from typing import Any, Iterator, Optional
@@ -70,7 +72,8 @@ class PoolWorker:
 
     def cancel(self) -> None:
         self.is_active_cancel = True  # signal for result callback
-        self.process.terminate()  # SIGTERM
+        # Use SIGUSR1 instead of SIGTERM to avoid conflicts with task signal handlers
+        os.kill(self.process.pid, signal.SIGUSR1)
 
     def get_data(self) -> dict[str, Any]:
         return {
--- a/dispatcher/worker/task.py
+++ b/dispatcher/worker/task.py
@@ -26,7 +26,9 @@ class WorkerSignalHandler:
     def __init__(self, worker_id):
         self.kill_now = False
         self.worker_id = worker_id
-        signal.signal(signal.SIGTERM, self.task_cancel)
+        # Use SIGUSR1 instead of SIGTERM for task cancellation
+        # This prevents conflicts with tasks that register their own SIGTERM handlers
+        signal.signal(signal.SIGUSR1, self.task_cancel)
         signal.signal(signal.SIGINT, self.exit_gracefully)
 
     def task_cancel(self, *args, **kwargs):
--- a/docs/design_notes.md
+++ b/docs/design_notes.md
@@ -82,6 +82,32 @@ In AWX dispatcher, a full queue may put messages into individual worker IPCs.
 This caused bad results, like delaying tasks due to long-running jobs,
 while the pool had many other workers free up in the mean time.
 
+
+### Worker Task Cancelation Signals
+
+Originally, the dispatcher library used **SIGTERM** to forcibly cancel a running task. However,
+some user code (or third-party libraries) register custom SIGTERM handlers, which caused
+collisions and made forcible timeouts/cancellations unreliable.
+
+To avoid these conflicts, dispatcher now uses **SIGUSR1** for per-task cancelation. This means:
+
+1. **Task**: When the dispatcher needs to cancel a running task (for example, due to a timeout),
+   it sends SIGUSR1 to that worker subprocess instead of SIGTERM.
+
+2. **Worker**: In the worker code (`WorkerSignalHandler`), the `SIGUSR1` handler raises
+   `DispatcherCancel`, which immediately stops the current task’s execution path.
+
+3. **Shutdown**: In contrast, a full worker shutdown or system-wide stop
+   can still use SIGTERM or SIGINT. That’s typically not per-task, but a broader
+   action telling the entire worker to exit.
+
+#### Why SIGUSR1
+
+SIGUSR1 is less commonly used by applications, so it’s less likely that user code
+will override it (unlike SIGTERM). This ensures that the dispatcher’s forced cancel
+mechanism remains reliable across a wide variety of user tasks, libraries, or Python
+frameworks that might expect to intercept SIGTERM for graceful shutdown.
+
 ### Notable python tasking systems
 
 https://taskiq-python.github.io/guide/architecture-overview.html#context
--- a/tests/integration/test_signal_handling.py
+++ b/tests/integration/test_signal_handling.py
@@ -1,0 +1,74 @@
+import asyncio
+import inspect
+import time
+
+import pytest
+
+from dispatcher.publish import task
+
+
+@task(queue="test_channel")
+def sigterm_interceptor_task():
+    """
+    This function is top-level in an importable module.
+    The dispatcher sees e.g. "tests.integration.test_signal_handling.sigterm_interceptor_task".
+    """
+    import signal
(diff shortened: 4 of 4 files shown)


dispatcher/service/pool.py
```diff
@@ -70,7 +72,8 @@ async def stop(self) -> None:
 
     def cancel(self) -> None:
         self.is_active_cancel = True  # signal for result callback
-        self.process.terminate()  # SIGTERM
+        # Use SIGUSR1 instead of SIGTERM to avoid conflicts with task signal handlers
+        os.kill(self.process.pid, signal.SIGUSR1)
```
AlanCoding, line 76: These linter errors take some getting used to.

```
dispatcher/service/pool.py:105: error: Argument 1 to "kill" has incompatible type "int | None"; expected "int"  [arg-type]
```

A process that is not running (or has never ran) will have `None` for a pid. You would never call `.cancel` for such a process, but the checker is not that smart. You can pre-check if pid is none and `return`, if it's not running then we'll say it's already canceled.

dispatcher/worker/task.py
```diff
@@ -26,7 +26,9 @@ class WorkerSignalHandler:
     def __init__(self, worker_id):
         self.kill_now = False
         self.worker_id = worker_id
-        signal.signal(signal.SIGTERM, self.task_cancel)
+        # Use SIGUSR1 instead of SIGTERM for task cancellation
+        # This prevents conflicts with tasks that register their own SIGTERM handlers
+        signal.signal(signal.SIGUSR1, self.task_cancel)
         signal.signal(signal.SIGINT, self.exit_gracefully)
```
AlanCoding, line 32: So SIGINT is Ctl+C

I had to look that up, and I think I just convinced myself that it should be kept in the current form. This could be deleted, or we could add another method to the worker stop method, but I'm honestly indifferent.

docs/design_notes.md
```diff
@@ -82,6 +82,32 @@ In AWX dispatcher, a full queue may put messages into individual worker IPCs.
...
+To avoid these conflicts, dispatcher now uses **SIGUSR1** for per-task cancelation. This means:
+
+1. **Task**: When the dispatcher needs to cancel a running task (for example, due to a timeout),
+   it sends SIGUSR1 to that worker subprocess instead of SIGTERM.
+
+2. **Worker**: In the worker code (`WorkerSignalHandler`), the `SIGUSR1` handler raises
+   `DispatcherCancel`, which immediately stops the current task’s execution path.
+
+3. **Shutdown**: In contrast, a full worker shutdown or system-wide stop
+   can still use SIGTERM or SIGINT. That’s typically not per-task, but a broader
```
AlanCoding, line 101: I don't think that's true about SIGTERM with the current code?

As a sustainability issue, if we do hijack SIGTERM (not saying that would be wrong), then the worker should probably re-connect it after finishing a task. Again, I feel like this is over-thinking. I'm just bringing it up because this one statement might technically be wrong by listing SIGTERM how you have it now.

dispatcher/service/pool.py
art-tapin: Will this work?

```python
self.is_active_cancel = True  # signal for result callback

pid = self.process.pid
   if pid is None:
       return 
       
os.kill(pid, signal.SIGUSR1)
```

Do I need to log it or can just simply return? I believe, at this stage it's already cancelled/not running?

tests/integration/test_signal_handling.py
```diff
@@ -0,0 +1,74 @@
...
+    node_id = canceled_info.pop("node_id", None)
+    assert node_id, "No node_id in reply"
+    assert canceled_info, f"Missing worker data: {canceled_info}"
+
+    # We'll search for any subdict that has 'uuid': our expected value
+    (worker_key, worker_data) = list(canceled_info.items())[0]
+    assert worker_data["uuid"] == uuid_val
+
+    # Wait for the dispatcher to declare all tasks cleared
+    await asyncio.wait_for(apg_dispatcher.pool.events.work_cleared.wait(), timeout=5)
```
AlanCoding, line 70: From a first look, this is completely duplicated with the above `clearing_task`. Because you didn't submit any new work between them - the only reason this isn't blocking is probably because you never called `.clear()` on the event, so you could await the `.wait()` all day long and it'll immediately continue.

tests/integration/test_signal_handling.py
```diff
@@ -0,0 +1,74 @@
...
+    a custom SIGTERM handler.
+    """
+
+    # Verify code implementation uses SIGUSR1
+    from dispatcher.service.pool import PoolWorker
+    from dispatcher.worker.task import WorkerSignalHandler
+    code_init = inspect.getsource(WorkerSignalHandler.__init__)
+    code_cancel = inspect.getsource(PoolWorker.cancel)
+    assert "SIGUSR1" in code_init
+    assert "SIGUSR1" in code_cancel
```
AlanCoding, line 41: I'm cool with this, but this block of code would best be broken into another test. It could live in the `tests/unit/` folder, because it has no dependency on the fixtures or other systems.

dispatcher/service/pool.py
```diff
@@ -70,7 +72,15 @@ async def stop(self) -> None:
 
     def cancel(self) -> None:
         self.is_active_cancel = True  # signal for result callback
-        self.process.terminate()  # SIGTERM
+
+        # If the process has never been started or is already gone, its pid may be None
+        pid = self.process.pid
+        if pid is None:
```
AlanCoding, line 78: I wish I was joking, but apparently MyPY is confused by the fact that you saved `pid` to a variable. using `if self.process.pid is None:` seems to fix it.

This is ridiculous, and I want to clarify I am _only_ cool with the type checking stuff because dispatcherd is lower-level than the other Django stuff, so it becomes achievable even with this stuff.

AlanCoding (approved):
Other than you fighting with the linter, this is in good enough shape. My comments are fairly minor.

dispatcher/worker/task.py
art-tapin: Leaving SIGINT handling unchanged for now; removed the old SIGTERM line as you suggested.

docs/design_notes.md
art-tapin: Reworded the design notes to avoid claiming SIGTERM is used for full shutdown. Now it mentions SIGINT or terminate().

tests/integration/test_signal_handling.py
art-tapin: Dropped the second clearing_task wait, as suggested :+1:

dispatcher/service/pool.py
```diff
@@ -70,7 +72,11 @@ async def stop(self) -> None:
 
     def cancel(self) -> None:
         self.is_active_cancel = True  # signal for result callback
-        self.process.terminate()  # SIGTERM
+
+        # If the process has never been started or is already gone, its pid may be None
+        if self.process.pid is None:
+            return  # it's effectively already canceled/not running
```
pb82, line 78: interesting, how is it possible that a process has no Pid?

pb82 (approved):
LGTM

Issue #61, opened 2025-02-10
A `dispatcherctl` entrypoint


AlanCoding:
This would help to make a demo of a fully complete dispatcher library.

This would work the same as `dispatcher-standalone` now, which we will probably rename. The ultimate idea is that we'll have 2 commands like:

 - `dispatcherd` - runs the service
 - `dispatcherctl` - interacts with the service

This would take the config file, complete with default lookup locations, the same as the service command. It would allow for sub-commands to do control-and-reply commands. In the ideal, it would default to local Unix socket communication, but this would need to be added as a broker first.

Alex-Izquierdo:
If we are going to use ansible-dispatcher, what would be the binaries? Would make sense to have a general entrypoint "ansible-dispatcher" with subcommands?

AlanCoding:
The idea is that this can parallel [supervisord](https://supervisord.org/) / `supervisorctl`. Or systemd / systemctl. I don't expect subcommands will make this easier. I see the CLI commands as for demo purposes now. I was even considering that we might only install the CLI entrypoints as an "extra", because they may be unwanted in production. However it's done, I do think it's important for a user-facing interface to control-and-reply commands for debugging and SOS reports. Linked patch.

PR #113, opened 2025-03-08
Add a `dispatcherctl` command, rework the demo with it

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/61

This renames `dispatcher-standalone` to `dispatcherd`, and then adds `dispatcherctl`.

I'm taking advantage of the new entrypoint to remove the 2-terminal-tab instructions from the manual demo. You can just to `docker logs` to see what the server is doing. The `run_demo.py` command doesn't seem to have any problem running against 2 servers as opposed to 1. Everything just happens in duplicate.

### DIFF
--- a/tools/demo/Dockerfile
+++ b/tools/demo/Dockerfile
@@ -1,0 +1,23 @@
+FROM ghcr.io/linuxcontainers/alpine:latest
+
+WORKDIR /dispatcherd
+
+RUN apk add --no-cache python3 py3-pip && ln -sf python3 /usr/bin/python
+
+RUN python3 -m venv /dispatcherd/venv
+
+RUN echo $pwd
+
+COPY . .
+
+RUN /dispatcherd/venv/bin/pip3 install -e .[pg_notify] --verbose
+
+COPY dispatcher.yml /dispatcherd/dispatcher1.yml
+COPY dispatcher.yml /dispatcherd/dispatcher2.yml
+
+RUN sed -i 's/\(node_id: \)demo-server-a/\1demo-server-1/' /dispatcherd/dispatcher1.yml
+RUN sed -i 's/\(node_id: \)demo-server-a/\1demo-server-2/' /dispatcherd/dispatcher2.yml
+
+RUN sed -i 's/\(host=\)localhost \(port=\)55777/\1msg_postgres \25432/' /dispatcherd/dispatcher1.yml
+RUN sed -i 's/\(host=\)localhost \(port=\)55777/\1msg_postgres \25432/' /dispatcherd/dispatcher2.yml
+
--- a/tools/demo/Dockerfile
+++ b/tools/demo/Dockerfile
@@ -8,9 +8,9 @@ RUN python3 -m venv /dispatcherd/venv
 
 RUN echo $pwd
 
-COPY . .
+COPY . ./src
 
-RUN /dispatcherd/venv/bin/pip3 install -e .[pg_notify] --verbose
+RUN /dispatcherd/venv/bin/pip3 install -e ./src[pg_notify] --verbose
 
 COPY dispatcher.yml /dispatcherd/dispatcher1.yml
 COPY dispatcher.yml /dispatcherd/dispatcher2.yml
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ linters:
 	mypy dispatcher
 
 demo:
-	docker compose up -d
+	docker compose up -d --wait
 
 stop-demo:
 	docker compose down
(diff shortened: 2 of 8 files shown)


tools/demo/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN apk add --no-cache python3 py3-pip && ln -sf python3 /usr/bin/python
+
+RUN python3 -m venv /dispatcherd/venv
+
+RUN echo $pwd
+
+COPY . .
+
+RUN /dispatcherd/venv/bin/pip3 install -e .[pg_notify] --verbose
```
AlanCoding, line 13: What would be better here would be to volume-mount the source so that you don't necessarily need to rebuild the image every time.

Talking this out as I think of things... that probably requires moving the instance-specific config files to a different location so they won't be in the volume mount. Like, right now, mounting the volume would make `/dispatcherd/dispatcher1.yml` go away. So in some way we may need a `src/` directory, different from where other misc stuff is put.

art-tapin: This is now implemented in the [d2552a1](https://github.com/ansible/dispatcherd/pull/113/commits/d2552a18a4f7ae03747eb8194734857befe25a77) right?

AlanCoding: Oh yes, I just pushed that in the last few minutes.

art-tapin (changes requested):
I ran:

```
grep -r -i "dispatcher-standalone" .
```
…and found leftover references in config.md and README.md
```
./docs/config.md:The provided entrypoint `dispatcher-standalone` ...
./docs/config.md:...the service (`dispatcher-standalone`) ...
./README.md:- A CLI entrypoint `dispatcher-standalone` ...
```
Could you replace those with **`dispatcherd`** to keep everything consistent? Something like:

- docs/config.md "`dispatcher-standalone`" → "`dispatcherd`"
- docs/config.md same for the mention of the service
- README.md "`dispatcher-standalone`" → "`dispatcherd`"

art-tapin (approved):
Hey Alan, overall, amazing work on this PR! Besides that minor cleanup needed for lingering “dispatcher-standalone” references, everything is really solid.

I ran several tests. For example, I did the following to check partial replies:

```bash
docker stop dispatcherd2
dispatcherctl workers --expected-replies=2
echo $?
```

Even though only demo-server-1 was fully active, the CLI still exited with 0 because the second node provided a minimal “last gasp” response instead of no response at all (maybe I didn’t give it enough time to completely wrap up). Other commands like `cancel` and `alive` worked exactly as advertised.

Really impressed with the polish – nicely done! I’m happy to approve this PR! Amazing job! 🥇

AlanCoding:
> Even though only demo-server-1 was fully active, the CLI still exited with 0 because the second node provided a minimal “last gasp” response instead of no response at all (maybe I didn’t give it enough time to completely wrap up). 

That was a good test case I didn't think of. I was interested, but it seems to give the expected exit code of 1.

```
$ dispatcherctl workers --expected-replies=2
DEBUG:dispatcher.cli:Configured standard out logging at DEBUG level
INFO:dispatcher.cli:Using config from file /home/alancoding/repos/dispatcher/dispatcher.yml
INFO:awx.main.dispatch.control:control-and-reply workers to None
INFO:dispatcher.brokers.pg_notify:Set up pg_notify listening on channel 'reply_to_e409f248_2ca7_460d_a6cc_c60bf904f291'
DEBUG:dispatcher.brokers.pg_notify:Sent pg_notify message of 83 chars to test_channel
DEBUG:dispatcher.brokers.pg_notify:Starting listening for pg_notify notifications
INFO:awx.main.dispatch.control:control-and-reply message returned in 1.0696587562561035 seconds
- node_id: demo-server-1
  worker-0:
    active_cancel: false
    age: 917.3461196170028
    current_task: null
    current_task_uuid: null
    finished_count: 306
    pid: 10
    status: ready
    worker_id: 0
  worker-1:
    active_cancel: false
    age: 917.3459000230068
    current_task: 'lambda: __import__("time").sleep(2)'
    current_task_uuid: sch-610
    finished_count: 304
    pid: 11
    status: ready
    worker_id: 1

ERROR:dispatcher.cli:Obtained only 1 of 2, exiting with non-zero code
(awx) alancoding@Sourdough:~/repos/dispatcher$ echo $?
1
```

We don't have coloring or anything so it's hard to see, but the last line does give an ERROR.

If we wanted color, AWX used `logutils`, but only for the dev environment.

https://github.com/ansible/awx/blob/devel/awx/main/utils/handlers.py#L111

PR #116, opened 2025-03-10
PoC Add metrics

chrismeyersfsu:
Something to have a conversation. This should not be merged. Showcases just in time metric generation.

### DIFF
--- a/dispatcher/service/main.py
+++ b/dispatcher/service/main.py
@@ -1,6 +1,17 @@
 import asyncio
 import json
 import logging
+import prometheus_client
+from prometheus_client import (
+    generate_latest,
+    Gauge,
+    Counter,
+    Enum,
+    CollectorRegistry,
+    parser,
+)
+from prometheus_client.core import GaugeMetricFamily, CounterMetricFamily, REGISTRY, StateSetMetricFamily
+from prometheus_client.registry import Collector
 import signal
 import time
 from typing import Iterable, Optional, Union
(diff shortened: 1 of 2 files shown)


dispatcher/service/main.py
```diff
@@ -1,6 +1,17 @@
 import asyncio
 import json
 import logging
+import prometheus_client
+from prometheus_client import (
```
AlanCoding, line 5: This would need to be a new dependency. It can be an optional "extra".

dispatcher/service/main.py
```diff
@@ -15,6 +26,53 @@
...
+        MetricsNamespace.__init__(self, namespace)
+        self._registry = registry
+
+        self._server = None
+        self._tid = None
+
+    def start(self):
+        try:
+            # TODO: addr for ipv6 ?
+            self._server, self._tid = prometheus_client.start_http_server(self.port(), addr='localhost', registry=self._registry)
```
AlanCoding, line 50: For documentation, this starts a thread

https://github.com/prometheus/client_python/blob/e3bfa1f10195b6959c5f49503762d07a47e1654c/prometheus_client/exposition.py#L238

AlanCoding:
I applied the following diff to prove out methods:

```diff
diff --git a/dispatcher/__init__.py b/dispatcher/__init__.py
index ae79987..8e9b351 100644
--- a/dispatcher/__init__.py
+++ b/dispatcher/__init__.py
@@ -13,6 +13,7 @@ def run_service() -> None:
     """
     loop = asyncio.get_event_loop()
     dispatcher = from_settings()
+    dispatcher.start_metrics()
     try:
         loop.run_until_complete(dispatcher.main())
     except KeyboardInterrupt:
diff --git a/dispatcher/service/main.py b/dispatcher/service/main.py
index c5d3de9..2d34939 100644
--- a/dispatcher/service/main.py
+++ b/dispatcher/service/main.py
@@ -287,8 +287,6 @@ class DispatcherMain:
 
     async def main(self) -> None:
         await self.connect_signals()
-        
-        self.start_metrics()
 
         try:
             await self.start_working()
```

It was not 100% obvious to me that this would work. It appears it does. Following your instructions, `curl http://localhost:8070` still gives apparently the same data. So this seems to work.

AlanCoding (changes requested):
This is a great first-pass, I learned a lot from it, and it is very cool.

My only problem is that it violates my intuitions of dispatcherd as a lower-level library. I'm also concerned about the implications for standardization, as I don't want dispatcherd to "own" the metrics severing structure. For next step I would propose the implementation:

 - add prometheus_client to the core dependencies in `pyproject.toml`
 - remove everything related to `start_http_server` here and move it to https://github.com/ansible/awx/compare/devel...feature_dispatcher
 - keep the `do_metrics` here as a callback for that

We would still be able to add some tests here, testing the callback (in a live context). And we could add further tests in AWX.

I think this would do well to maximize our future options for rolling out standard metrics and configuration of it.

AlanCoding:
Closing, with a replacement up.

PR #118, opened 2025-03-11
Fix missing `await`

AlanCoding:
Seen in test logs that method was never awaited. This was supposed to be an optimization to prevent starting the thread in some cases, but that optimization was not working, so it was a subtle bug.


art-tapin (approved):
I've reviewed the changes in the PR, test are green. Overall, the changes look good.

LGTM! 👍

PR #119, opened 2025-03-11
[unfinished] Add scaler class

AlanCoding:
This needs https://github.com/ansible/dispatcherd/pull/111 before it can move ahead. After a painful rebase this can be finished.

The goal of this and 111 is to reduce the total cognitive load of `WorkerPool`.

At this checkpoint, `WorkerPool` winds up only managing the results reading task by itself. Everything else is somehow managed by a related class/object. Indeed, the class might have more logic in scaling that it has other logic. Predictably, you wind up passing _a lot_ of data. But you still get some organizational benefits - like the fact that `WorkerPool` doesn't have any direct reason to count the running tasks.

The config implications are not finished. Right now `max_workers` are stored in `pool_kwargs`. This change would imply they are moved to `scalar_kwargs` which would be a breaking change (also not finished yet). This is so disruptive that I might opt to put some minimal translation logic in, maybe upgrade to `"version": 3`.

So for now I just want to disclose the refactoring direction I'm headed in, ping @art-tapin @pb82

### DIFF
--- a/dispatcher/service/pool.py
+++ b/dispatcher/service/pool.py
@@ -10,6 +10,7 @@ from .blocker import Blocker
 from .next_wakeup_runner import HasWakeup, NextWakeupRunner
 from .process import ProcessManager, ProcessProxy
 from .queuer import Queuer
+from .scaler import Scaler, ScalerConfig
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


AlanCoding:
I want to get https://github.com/ansible/dispatcherd/pull/121 in before I continue with this, marking blocked.

PR #120, opened 2025-03-11
Detect and clean up unexpectedly terminated worker processes

### LINKED ISSUE #97: Detect and remove workers whose process is not alive
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

This asks that we port over AWX functionality to detect when a worker's process is dead.

https://github.com/ansible/awx/blob/7b8b37d9a8e46b3cda101f3c5691c870b9de46da/awx/main/dispatch/pool.py#L388-L405

I would ask that we do _not_ port over this part:

```python
                        try:
                            for j in UnifiedJob.objects.filter(celery_task_id=w.current_task['uuid']):
                                reaper.reap_job(j, 'failed')
                        except Exception:
                            logger.exception('failed to reap job UUID {}'.format(w.current_task['uuid']))
```

Because I believe this is unnecessary. This can be done (and maybe is already done) in logic elsewhere based on the database `Job` data including `status` and `work_unit_id`.

So we're only looking for cleanup in the class that another process issues a SIGKILL to a worker process. If another process issues a SIGTERM, the worker may exit normally and send a message to the parent indicating this. In that case, current logic might actually work, but we could use testing of this.

### Steps to reproduce

Again, we have a clear test case that we can write in current test suite.

1. Spawn the dispatcher in integration tests with like 2 workers in the "ready" state
2. Run the "worker" control-and-reply command to get worker data
3. SIGKILL one of the pids obtained from the response from (2)
4. Kick the method on the dispatcher that does the cleanup
5. Assert that we have 1 fewer workers in the "ready" state, showing that the lost process was detected

I do not expect this would work right now in any sense.

### Current results

_No response_

### Sugested feature result

detects dead processes in pool.

### Additional information

_No response_

art-tapin:
Fixes #97
JIRA: [AAP-41124](https://issues.redhat.com/browse/AAP-41124)

Added logic to detect when a worker process has terminated unexpectedly (like from a SIGKILL or crash) without properly notifying the parent process. These workers are now automatically marked with error status, their tasks are counted as canceled, and they're removed from the pool after the standard removal wait period.

This prevents "zombie" workers from cluttering the pool and tasks from appearing to run forever when their worker has died.

## Testing:
**Prerequisites**
* Standard test setup (pytest with dispatcher dependencies)

**Steps to test**
1. Run: `python -m pytest  tests/unit/service/test_worker_liveness.py`
2. Verify all tests pass, showing that dead workers are properly detected, marked, and removed

## Notes:
Rationale behind modify the `manage_old_workers()` method in pool.py:

1. This method already handles the worker lifecycle - it was a natural extension to add detection of unexpected terminations
2. The periodic checking approach integrates seamlessly with the existing architecture
3. No changes were needed to the worker code itself - the detection happens entirely in the parent process

The implementation usess the existing worker state model, creating a clean pathway from detection (setting status to 'error') through to eventual removal. By incrementing the canceled_count, we ensure accurate task accounting, which was previously missing for tasks interrupted by worker termination.

### DIFF
--- a/dispatcher/service/pool.py
+++ b/dispatcher/service/pool.py
@@ -298,12 +298,24 @@ class WorkerPool:
         """
         remove_ids = []
         for worker in self.workers.values():
+            # Check for workers that died unexpectedly
+            if worker.status not in ['retired', 'error', 'exited'] and not worker.process.is_alive():
+                logger.error(f'Worker {worker.worker_id} pid={worker.process.pid} has died unexpectedly, status was {worker.status}')
+
+                if worker.current_task:
+                    uuid = worker.current_task.get('uuid', '<unknown>')
+                    logger.error(f'Task (uuid={uuid}) was running on worker {worker.worker_id} but the worker died unexpectedly')
+                    self.canceled_count += 1
+                    worker.is_active_cancel = False  # Ensure it's not processed by timeout runner
+
+                worker.status = 'error'
+                worker.retired_at = time.monotonic()
+
             if worker.status == 'exited':
                 await worker.stop()  # happy path
             elif worker.status == 'stopping' and worker.stopping_at and (time.monotonic() - worker.stopping_at) > self.worker_stop_wait:
                 logger.warning(f'Worker id={worker.worker_id} failed to respond to stop signal')
                 await worker.stop()  # agressively bring down process
-
             elif worker.status in ['retired', 'error'] and worker.retired_at and (time.monotonic() - worker.retired_at) > self.worker_removal_wait:
                 remove_ids.append(worker.worker_id)
 
--- a/dispatcher/service/pool.py
+++ b/dispatcher/service/pool.py
@@ -299,7 +299,7 @@ class WorkerPool:
         remove_ids = []
         for worker in self.workers.values():
             # Check for workers that died unexpectedly
-            if worker.status not in ['retired', 'error', 'exited'] and not worker.process.is_alive():
+            if worker.status not in ['retired', 'error', 'exited', 'initialized', 'spawned'] and not worker.process.is_alive():
                 logger.error(f'Worker {worker.worker_id} pid={worker.process.pid} has died unexpectedly, status was {worker.status}')
 
                 if worker.current_task:
--- a/tests/unit/service/test_worker_liveness.py
+++ b/tests/unit/service/test_worker_liveness.py
@@ -11,29 +11,38 @@ from dispatcher.service.process import ProcessManager
 @pytest.mark.asyncio
 async def test_detect_unexpectedly_dead_worker(test_settings, caplog):
     """
-    Verify that when a worker that is processing a task dies unexpectedly,
-    it is marked with an error status, its task is canceled, and proper logging occurs.
+    Verify that if a worker dies unexpectedly while processing a task,
+    it is marked as 'error', its task is canceled, and appropriate error logs are produced.
     """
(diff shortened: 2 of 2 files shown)


tests/unit/service/test_worker_liveness.py
```diff
@@ -0,0 +1,82 @@
...
+    pm = ProcessManager(settings=test_settings)
+    pool = WorkerPool(pm, min_workers=2, max_workers=5)
+    await pool.up()
+    worker = pool.workers[0]
+    worker.status = 'ready'
+    worker.current_task = {'uuid': 'test-task-123'}
+
+    # Simulate unexpected process death by forcing is_alive to return False.
+    with caplog.at_level("DEBUG"):
+        with mock.patch.object(worker.process, 'is_alive', return_value=False):
```
AlanCoding, line 27: hmm... honestly, my preference would be that you just outright call `worker.process.kill()`. Literally kill the process. IIRC, this does a SIGKILL, but check docs. If you do a SIGTERM, it may still be possible that the worker sends the exit message, which won't hit your intended code path.

tests/unit/service/test_worker_liveness.py
```diff
@@ -0,0 +1,82 @@
...
+async def test_dead_worker_with_no_task(test_settings):
+    """
+    Verify that a dead worker which is not running any task is marked as error,
+    without increasing the canceled task count.
+    """
+    pm = ProcessManager(settings=test_settings)
+    pool = WorkerPool(pm, min_workers=1, max_workers=3)
+    await pool.up()
+    worker = pool.workers[0]
+    worker.status = 'ready'
```
AlanCoding, line 71: I think checking that status is "ready" might be racy, because `.up()` doesn't guarantee that status. It only gets that status after the worker sends its ready message. So we might want to add an `await` for, IIRC the `workers_ready` event. This should work because you're only going up to 1 worker.

dispatcher/service/pool.py
```diff
@@ -290,12 +290,24 @@ async def manage_old_workers(self) -> None:
         """
         remove_ids = []
         for worker in self.workers.values():
+            # Check for workers that died unexpectedly
+            if worker.status not in ['retired', 'error', 'exited'] and not worker.process.is_alive():
```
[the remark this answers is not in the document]
art-tapin: Good catch! Added both states to the exclusion list. We definitely don't want to trigger false positives when workers are still starting up :+1:

tests/unit/service/test_worker_liveness.py
art-tapin: Thanks! Now it uses worker.process.kill(). Also improved the cleanup by overriding the cancel method for shutdown - that fixes the ProcessLookupError that I had before.

tests/unit/service/test_worker_liveness.py
art-tapin: Hey, I tried using the workers_ready event as suggested, but it caused an infinite loop in this test. I ended up setting the status manually instead. Is that okay?

AlanCoding (approved):

Issue #71, opened 2025-02-11
Add features to `bind=True` object: control tasks & start tasks


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

In my current patch to add `bind` as a task option, it is only providing static data for the moment.

This requests that we add 2 new methods to the object, these are dynamic, and interact with queues available to worker code:
1. Run a control task via the IPC mechanism and get a response over the IPC mechanism
2. Start another task, do not get a reply

### Steps to reproduce

N/A feature request

### Sugested feature result

aforementioned callable features.

### Additional information

Not in scope right now: async versions

A keep-alive method.

May file issue for these later.

PR #121, opened 2025-03-12
More type hints & protocols, refactor `control.py` to be broker-based

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/109

Fixes https://github.com/ansible/dispatcherd/issues/124

Confirmed that the demo still looks good with this change.

This does make structural changes to how control-and-reply works, but I think that is all for the better.

PR #122, opened 2025-03-12
Allow running control tasks from worker

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/71

I just want to demonstrate the gross things so other people don't feel bad doing them, in case someone else goes to pick this up.

From logs running this new test:

```
DEBUG:dispatcher.service.queuer:Dispatching task (uuid=b00b237a-d7bb-4bb0-97e1-98cd0050f1bf) to worker (id=0)
DEBUG:dispatcher.service.pool:Dispatching task (uuid=b00b237a-d7bb-4bb0-97e1-98cd0050f1bf) to worker (id=0)
DEBUG:dispatcher.worker.task:task (uuid=b00b237a-d7bb-4bb0-97e1-98cd0050f1bf) starting tests.data.methods.prints_running_tasks(*[]) on worker 0
{'worker-0': {'task': 'tests.data.methods.prints_running_tasks', 'time_pub': 1741794622.4742262, 'uuid': 'b00b237a-d7bb-4bb0-97e1-98cd0050f1bf', 'args': [], 'kwargs': {}, 'channel': 'test_channel'}}
DEBUG:dispatcher.service.pool:Worker 0 finished task (uuid=b00b237a-d7bb-4bb0-97e1-98cd0050f1bf), ct=0
```

So that's it, that's what we need.

Clearly, the interfaces could use some polishing. See https://github.com/ansible/dispatcherd/pull/121. Here, we're passing some unnecessary data since we passed a whole `DispatcherMain` anyway. But we should prioritize this as a feature before making it completely pretty.

### DIFF
--- a/dispatcherd/protocols.py
+++ b/dispatcherd/protocols.py
@@ -164,7 +164,7 @@ class WorkerPool(Protocol):
     queuer: Queuer
     blocker: Blocker
 
-    async def start_working(self, forking_lock: asyncio.Lock, exit_event: Optional[asyncio.Event] = None) -> None:
+    async def start_working(self, dispatcher: 'DispatcherMain', exit_event: Optional[asyncio.Event] = None) -> None:
         """Start persistent asyncio tasks, including asychronously starting worker subprocesses"""
         ...
 
@@ -186,7 +186,7 @@ class DispatcherMain(Protocol):
 
     pool: WorkerPool
     delayed_messages: set
-    lock_fd: asyncio.Lock  # Forking and locking may need to be serialized, which this does
+    fd_lock: asyncio.Lock  # Forking and locking may need to be serialized, which this does
 
     async def main(self) -> None:
         """This is the method that runs the service, bring your own event loop"""
--- a/dispatcherd/protocols.py
+++ b/dispatcherd/protocols.py
@@ -186,6 +186,7 @@ class DispatcherMain(Protocol):
 
     pool: WorkerPool
     delayed_messages: set
+    lock_fd: asyncio.Lock  # Forking and locking may need to be serialized, which this does
 
     async def main(self) -> None:
         """This is the method that runs the service, bring your own event loop"""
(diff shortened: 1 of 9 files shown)


AlanCoding:
It may be unreasonable to get the linters to pass before 121 is merged. As I learned there, `WorkerPool` doesn't actually conform to its protocol, and it's not feasible to do so when that protocol declares a `dict` to be part of the spec, due to annoying philosophical reasons.

art-tapin (approved):
Hey @AlanCoding, this PR looks solid! I tested it by running a script that calls `prints_running_tasks` which uses the new `binder.control()` method. The logs clearly show the full cycle working - task dispatch → control command execution → result returned to worker. Very clean implementation.

A few suggestions that would make this even better:

- Logging - I really missed something like "Processing control command X from worker Y" and I think would be a lifesaver when diagnosing issues

- Documentation - i think, it deserves more documentation with examples. Maybe some sample use cases like "checking if another task is running" or "making decisions based on worker status".

- What about 'start tasks' part? - this PR handles the control tasks part of #71 perfectly. In the issue description you mentioned that issue also requested ability to start other tasks. Are you planning to address that in a future PR or did I miss something? (you mentioned 'Fixes issue 71` in the PR description)
> "This requests that we add 2 new methods to the object... Run a control task via the IPC mechanism and get a response over the IPC mechanism; Start another task, do not get a reply"

✅ LGTM

:1st_place_medal:

AlanCoding:
> What about 'start tasks' part?

This would allow you to run a control task from the worker, but https://github.com/ansible/dispatcherd/pull/132 would allow control tasks to start tasks. That way, there's no additional mechanism just for this purpose. There is still the issue of _disabling_ this via the config for security-ish purposes, which https://github.com/ansible/dispatcherd/issues/138 touches on, but these need to be enumerated more fully. Both running control tasks from the worker and starting tasks via control tasks should be thing that can be turned off.

AlanCoding:
I agree docs could be better, but for now I'm going to look at merging this and putting that work in https://github.com/ansible/dispatcherd/pull/132 after the rebase.

Issue #100, opened 2025-02-28
Develop solution for pg_notify character limit


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Bug Summary

I want to rely more-and-more on control-and-reply messages. With asyncio, this has become much more appealing as they are so much faster.

However, this has a character limit, like 8,000. And at this point, I think the easiest thing is to just solve the problem.

If a message is over that amount, we should just break it up into smaller messages. This is somewhat obviously possible. This could be implemented in `pg_notify.py` module (by moving the JSON parsing there) or in `DispatcherMain` with some other utility files involved.

### Environment

N/A

### Steps to reproduce

Do a method call with `args` or `kwargs` that parse to >8,000 characters.

It should error as things are.

### Actual results

This is actually worse when the "reply" message is big, because then replies are dropped with the listener not knowing what happened.

### Expected results

For this, measure the size of the dumped JSON to send, and if it's over the limit, break it up into chunks and wrap those chunks in more JSON. When receiving, save messages in memory until fully combined.

You could modify the specifics, but this is the general idea.

AlanCoding:
Related, I have https://github.com/ansible/dispatcherd/pull/113 up which adds `dispatcherctl` as an entrypoint but it still goes through pg_notify now. We would combine this with a "socket" broker to effectively solve this problem.

But I'd argue that control messages should still work with pg_notify if they go past the character limit, and I think it would be best to just solve this ping @chrismeyersfsu

PR #125, opened 2025-03-13
Add Unix socket broker

AlanCoding:
This is maybe, perhaps, a solution to https://github.com/ansible/dispatcherd/issues/100 but it is debatable.

However, it was part of the vision of `dispatcherctl` described in https://github.com/ansible/dispatcherd/issues/61

This was more complicated than I expected, but I got it working.

Using the "old" demo with 2 terminal tabs.

```
# tab 1
PYTHONPATH=. dispatcherd
# tab 2
./run_demo.py
```

It now get a bit awkward because you can no longer get a response from multiple servers. But we get to the end of the demo, and it prints everything. In my first pass it broke half-way through because the message was longer than 1024. I think I'm still putting in "reply_to" which the server just ignores.

Protocols need to be updated and basically everything aside from making it run the demo once, which is all it does right now.

### DIFF
--- a/dispatcherd/brokers/socket.py
+++ b/dispatcherd/brokers/socket.py
@@ -1,0 +1,223 @@
+import asyncio
+import json
+import logging
+import os
+import socket
+from typing import Any, AsyncGenerator, Callable, Coroutine, Iterator, Optional, Union
+
+from ..protocols import Broker as BrokerProtocol
+
+logger = logging.getLogger(__name__)
+
+
+class Client:
+    def __init__(self, client_id: int, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
+        self.client_id = client_id
+        self.reader = reader
+        self.writer = writer
+        self.listen_loop_active = False
+        # This is needed for task management betewen the client tasks and the main aprocess_notify
+        # if the client task starts listening, then we can not send replies
+        # so this waits for the caller method to add replies to stack before continuing
+        self.yield_clear = asyncio.Event()
+        self.replies_to_send: list = []
+
+    def write(self, message: str, /) -> None:
+        self.writer.write((message + '\n').encode())
+
+    def queue_reply(self, reply: str, /) -> None:
+        self.replies_to_send.append(reply)
+
+    async def send_replies(self) -> None:
+        for reply in self.replies_to_send.copy():
+            logger.info(f'Sending reply to client_id={self.client_id} len={len(reply)}')
+            self.write(reply)
+        else:
+            logger.info(f'No replies to send to client_id={self.client_id}')
+        await self.writer.drain()
+        self.replies_to_send = []
+
+
+def extract_json(message: str) -> Iterator[str]:
+    """With message that may be an incomplete JSON string, yield JSON-complete strings and leftover"""
+    decoder = json.JSONDecoder()
+    pos = 0
+    length = len(message)
+    while pos < length:
+        try:
+            _, index = decoder.raw_decode(message, pos)
+            json_msg = message[pos:index]
+            yield json_msg
+            pos = index
+        except json.JSONDecodeError:
+            break
+
+
+class Broker(BrokerProtocol):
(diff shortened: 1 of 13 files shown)


AlanCoding:
Oh, here's an interesting find from the demo output:

```
  Task-234 is done=False
   trace:
Stack for <Task pending name='Task-234' coro=<Broker._add_client() running at /home/alancoding/repos/dispatcher/dispatcher/brokers/socket.py:45> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[StreamReaderProtocol.connection_made.<locals>.callback() at /usr/lib64/python3.12/asyncio/streams.py:248]> (most recent call last):
  File "/home/alancoding/repos/dispatcher/dispatcher/brokers/socket.py", line 45, in _add_client
    line = await client.reader.readline()
```

It's right about where we expected it to be.

AlanCoding:
Again, I hit an issue where it will be difficult to move forward before https://github.com/ansible/dispatcherd/pull/121 lands.

When running commands as `dispatcherctl` entrypoint (the ultimate point of this), it can hang waiting for a reply. (yes this means timeout isn't working)

The reply isn't coming because `origin` information is 0. This is, predictably, because the `control.py` module has not had its contract updated to the same as `DispatcherMain`. However, the PR 121 removes the need to copy this contract in the first place. So that's where this leaves us.

AlanCoding:
I figured out what it was. In asyncio, there are problems having parallel tasks both listening and writing to the same connection. This is roughly the same issue I hit with pg_notify. The solution is same in form, except for the multiple tasks for different clients.

dispatcher/brokers/socket.py
```diff
@@ -0,0 +1,205 @@
...
+
+                if connected_callback:
+                    connected_callback()
+
+                received_ct = 0
+                buffer = ''
+                while True:
+                    response = sock.recv(1024).decode().strip()
+
+                    if response.endswith('}'):
```
AlanCoding, line 180: this is flawed, I'm going to work on a solution for this.

chrismeyersfsu (approved):

dispatcherd/brokers/socket.py
```diff
@@ -0,0 +1,223 @@
...
+                sock.settimeout(timeout)
+                sock.connect(self.socket_path)
+
+                if connected_callback:
+                    connected_callback()
+
+                received_ct = 0
+                buffer = ''
+                while True:
+                    response = sock.recv(1024).decode().strip()
```
chrismeyersfsu, line 194: I can confirm, if you bork the server, the timeout associated with the socket _does_ trigger.

:+1: 

```
  File "/home/cmeyers/github.com/ansible/dispatcherd/.venv/bin/dispatcherctl", line 8, in <module>
    sys.exit(control())
             ~~~~~~~^^
  File "/home/cmeyers/github.com/ansible/dispatcherd/dispatcherd/cli.py", line 84, in control
    returned = ctl.control_with_reply(args.command, data=data, expected_replies=args.expected_replies)
  File "/home/cmeyers/github.com/ansible/dispatcherd/dispatcherd/control.py", line 111, in control_with_reply
    for channel, payload in broker.process_notify(connected_callback=connected_callback, max_messages=expected_replies, timeout=timeout):
                            ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cmeyers/github.com/ansible/dispatcherd/dispatcherd/brokers/socket.py", line 195, in process_notify
    response = sock.recv(1024).decode().strip()
               ~~~~~~~~~^^^^^^
TimeoutError: timed out
```

AlanCoding:
I vetted the basic functionality with AWX and it's working so I'm going ahead with merge. Some code organizational issues are documented, but this isn't meant to be polished at this point.

Issue #2, opened 2024-10-17
Broker self-checks


AlanCoding:
As a part of this, we provably need:

https://github.com/ansible/awx/pull/14749

Evidence from the field and experiments have confirmed, ultimately, there is no other option. That has to be implemented.

However, doing a full self-check by passing messages is costly. It is possible we could have a fast self-check and a slow self-check. The fast self-check would probably be what is described here, which is what AWX community members have proposed before:

https://www.psycopg.org/psycopg3/docs/advanced/async.html

```python
while True:
    try:
        await asyncio.wait_for(ev.wait(), 60.0)
    except asyncio.TimeoutError:
        continue  # No FD activity detected in one minute

    # Activity detected. Is the connection still ok?
    try:
        await conn.execute("SELECT 1")
    except psycopg.OperationalError:
        # Guess what happened
        ...
```

Is it necessary to have 2 types of checks? Maybe not, I created this issue to document options.

AlanCoding:
I'm going to propose that we _initiate_ this from the _main_ thread. That is, replace this:

https://github.com/ansible/dispatcher/blob/ac485e25cfaf32117cd6807af5abca3264496831/dispatcher/service/main.py#L228

Instead of waiting for the event forever, let's add a timeout on the wait, and put it in a `while True:` loop. If that `await` _times out_ then we _do not exit_, but if it does _not time out_ that means the exit event was set, and we _do_ want to `break` out of the forever loop.

When a timeout happens, that means we're due for a check. Because we're writing async code, this will require a tick-tock kind of pattern. I will propose we add methods to the `Broker` protocol:

1. `async def initiate_self_check`
2. `async def get_self_check_result`

On every timeout, we would check if we got a bad result from last time, calling (2). If it's never been initiated, then that's probably due to startup, do nothing. If a check result is good, then do nothing, if the check result was bad, then go into `DispatcherMain` code to re-start the broker and mark a strike against it.

After that, we need to initiate a check, (1). This will follow the method I described in AWX. In the `Broker` class, we need to implement publishing a message to a self-check queue. The self-check queue needs to be defined in settings.

The message needs to have a special format / designation. When the message is received, (either in `Broker` or `DispatcherMain`, undecided), then we set a flag saying that the check passed. The message needs to contain the `node_id` and this needs to match the local `node_id` so that we don't say it checked due to messages from other nodes.

AlanCoding:
Documenting 2 test cases:

[...]

PR #126, opened 2025-03-14
[WIP] feat: detect broken connection and reconnect broker

pb82:
Fixes https://github.com/ansible/dispatcherd/issues/2

This PR implements periodic broker self checks. If a broken pg_notify connection is detected, it will attempt to reconnect.

TODO: tests, sanity checking

### DIFF
--- a/dispatcher/brokers/pg_notify.py
+++ b/dispatcher/brokers/pg_notify.py
@@ -79,7 +79,10 @@ class Broker:
 
         # Make sure that the self check channel is always present
         if not self.SELF_CHECK_CHANNEL in channels:
-            channels.append(self.SELF_CHECK_CHANNEL)
+            if isinstance(channels, list):
+                channels.append(self.SELF_CHECK_CHANNEL)
+            elif isinstance(channels, tuple):
+                channels = channels + (self.SELF_CHECK_CHANNEL,)
 
         self.channels = channels
         self.default_publish_channel = default_publish_channel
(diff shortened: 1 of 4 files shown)


dispatcher/service/main.py
```diff
@@ -234,7 +236,31 @@ async def main(self) -> None:
...
+                try:
+                    await asyncio.wait_for(self.events.exit_event.wait(), 5.0)
+
+                    # exit_event has fired, process should exit
+                    break
+                except asyncio.TimeoutError:
+                    logger.info(f'initiating broker self check for node-id {self.node_id}')
+                    for producer in self.producers:
+                        if isinstance(producer, BrokeredProducer):
+                            result = await producer.broker.get_self_check_result(self.node_id)
```
AlanCoding, line 250: What's actually super non-obvious is that it might not be okay to run 2 listening tasks at the same time. There are actually a lot of hazards with connection management in async code, and every I've fought this I've given up and went back to always forever only having a single task interact with psycopg.

Separate from that, this wouldn't have a timeout anyway, which is the failure scenario.

dispatcher/brokers/pg_notify.py
```diff
@@ -33,6 +36,7 @@ def create_connection(**config) -> psycopg.Connection:
 
 class Broker:
     NOTIFY_QUERY_TEMPLATE = 'SELECT pg_notify(%s, %s);'
+    SELF_CHECK_CHANNEL = 'pgnotify_self_check'
```
AlanCoding, line 39: This would result in all nodes sending their health check messages to all connected nodes. This isn't a good idea for scaling, but also wouldn't work unless there was some other identifying information in the message it used to verify that it came from itself. So this could be addressed by adding a uuid to it, and adding some expected uuid to the message might be a good idea anyway to rule out any possibility of getting wires crossed.

AlanCoding:
I may have failed to fully digest the consequences of https://github.com/psycopg/psycopg/issues/340 myself, and this requires some changes to the methodology to solve this problem.

From some light research, we must have the wakeup for _sending_ the message handled by the broker. Only viable option may be the `timeout` kwarg on the async version of `.notifies` from psycopg.

https://www.psycopg.org/psycopg3/docs/api/connections.html#psycopg.AsyncConnection.notifies

To set the stage, our next obvious alternative would involve manually calling `.__anext__`. That would give more options to weave together timing of the main loop with the broker listening loop. I don't like it, but the choice is still unsettled.

Issue #127, opened 2025-03-17
Switch test server from `pytest-asyncio` to subprocess


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

Filing partially to replace https://github.com/ansible/dispatcherd/issues/81

Taking lessons I learned from https://github.com/ansible/dispatcherd/pull/125, I want to describe steps for finally fixing the integration test organization. I'll outline initially in a numbered list:

1. Remove `acontrol_with_reply` and everything associated with this type of use. Only synchronous clients will be supported.
2. Fixtures such as `apg_dispatcher` will be converted to a synchronous version, using fixture from https://github.com/ansible/dispatcherd/issues/9. This will have to grow a couple of communication utilities and can be merged with the WIP of benchmark tests. In particular, we will probably do socket communication for a select set of asyncio events. Obviously, the subprocess will listen for an exit message.
3. All asyncio full-server tests will be converted to be synchronous tests
4. Wherever possible, the server fixtures will be changed to class or module scope to make tests faster for all non-destructive cases

The test requirement `pytest-asyncio` may still remain for some niche unit tests.

### Steps to reproduce

testing work

### Sugested feature result

Integration tests are all synchronous code.

The definition of "integration" will be that it uses a full server in a subprocess.

The full-server fixture will be shared among several tests whenever possible.

As per related issues, utilities to enable this will be in a `testing` submodule inside the main import path.

AlanCoding:
I didn't fully wrap this up with the "why?"

Because, writing `async` tests mean that we write a duplicate version of all _client_ functions in asyncio code. While this might eventually be a feature request, we have no immediate need for it in any non-testing sense. Even if we intend to do this, the current protocol for `Broker` classes are not fully coherent, as writing the socket client has revealed to me. In the pg-notify broker, we use the same methods for the client control-with-reply as for the server. This has an unwritten assumption that the broker behaves symmetrically (client listening has the same form as server-side listening). This is not true for the socket broker, and many other types of brokers.

[...]

Issue #9, opened 2024-10-23
Add `dispatcher/testing/` utilities to help testing


AlanCoding:
A good use case for this is AWX which has the messaging system disabled in tests.

We should add some `@pytest.fixture` methods that serve various off-the-shelf methods that other projects can use. Specifically I have in mind:
 - a context manager to ignore messages for duration it is active, what AWX does now, effectively.
 - a context manager to run a thread in the background, that listens to a `multiprocessing.Queue`, as its own kind of broker.

This would be a similar structure to:

https://github.com/ansible/django-ansible-base/tree/devel/ansible_base/lib/testing

PR #129, opened 2025-03-18
Start running synchronous tests, separate from asyncio tests

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/9

Fixes https://github.com/ansible/dispatcherd/issues/127

I described in more recent issues - the intent would be to ultimately replace all the existing async tests.

This has a nasty warning:

```
/home/alancoding/venvs/awx/lib64/python3.12/site-packages/psycopg/_connection_base.py:146: ResourceWarning: connection <psycopg.Connection [IDLE] (host=localhost port=55777 user=dispatch database=dispatch_db) at 0x7efbef973710> was deleted while still open. Please use 'with' or '.close()' to close the connection
```

I expect that will be resolved with connection closing @art-tapin did in https://github.com/ansible/dispatcherd/pull/121. Maybe.

Right now this test is slow, but the idea is that we'll start to reuse the fixture between tests and then it will start to go faster. The idea is also to reuse this in AWX and eda-server.

### DIFF
--- a/dispatcherd/service/process.py
+++ b/dispatcherd/service/process.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import logging
 import multiprocessing
 from multiprocessing.context import BaseContext
 from types import ModuleType
@@ -8,6 +9,8 @@ from typing import Any, Callable, Iterable, Optional, Union
 from ..config import LazySettings
 from ..config import settings as global_settings
 from ..worker.task import work_loop
+
+logger = logging.getLogger(__name__)
 
 
 class ProcessProxy:
@@ -75,7 +78,7 @@ class ProcessManager:
 
     def __init__(self, settings: LazySettings = global_settings) -> None:
         self.ctx = multiprocessing.get_context(self.mp_context)
-        self.finished_queue: multiprocessing.Queue = self.ctx.Queue()
+        self._finished_queue: Optional[multiprocessing.Queue] = self.ctx.Queue()
 
         # Settings will be passed to the workers to initialize dispatcher settings
         settings_config: dict = settings.serialize()
@@ -85,6 +88,14 @@ class ProcessManager:
         self.settings_stash: str = json.dumps(settings_config)
 
         self._loop: Optional[asyncio.AbstractEventLoop] = None
+
+    @property
+    def finished_queue(self) -> multiprocessing.Queue:
+        """The processor manager owns the lifecycle of the finished queue, so if pool is restarted we have to handle this"""
+        if self._finished_queue:
+            return self._finished_queue
+        self._finished_queue = self.ctx.Queue()
+        return self._finished_queue
 
     def get_event_loop(self) -> asyncio.AbstractEventLoop:
         if self._loop:
@@ -107,6 +118,12 @@ class ProcessManager:
         message = await self.get_event_loop().run_in_executor(None, self.finished_queue.get)
         return message
 
+    def shutdown(self) -> None:
+        if self._finished_queue:
+            logger.debug('Closing finished queue')
+            self._finished_queue.close()
+            self._finished_queue = None
+
 
 class ForkServerManager(ProcessManager):
     mp_context = 'forkserver'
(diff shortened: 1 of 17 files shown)


AlanCoding:
Tests are timeouts.

I suspect, though I do not know, this might be because I didn't call `process.close()`. It is also the first time we ran both subprocess and asyncio tests in the same run, so it may also not be surprising that we hit new hangs in this situation. But it's hard to clean up processes.

dispatcherd/service/process.py
```diff
@@ -99,10 +99,17 @@ async def read_finished(self) -> dict[str, Union[str, int]]:
         message = await self.get_event_loop().run_in_executor(None, self.finished_queue.get)
         return message
 
+    def shutdown(self) -> None:
+        self.finished_queue.close()
```
AlanCoding, line 103: Tests are working but I'm seeing an order-of-operations error.

```
ERROR:dispatcherd.service.main:Pool manager encountered error
  File "/home/alancoding/repos/dispatcher/dispatcherd/service/main.py", line 103, in shutdown
    await self.pool.shutdown()
  File "/home/alancoding/repos/dispatcher/dispatcherd/service/pool.py", line 454, in shutdown
    self.process_manager.finished_queue.put('stop')
  File "/usr/lib64/python3.12/multiprocessing/queues.py", line 88, in put
    raise ValueError(f"Queue {self!r} is closed")
ValueError: Queue <multiprocessing.queues.Queue object at 0x7f4cd2fa7e30> is closed
```

Shouldn't be too hard to figure out.

AlanCoding:
It turns out that this works:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index cf5dcc3..ff77dbc 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,4 +65,3 @@ pg_notify = ["psycopg[binary]>=3.2.4"]
 
 [tool.pytest.ini_options]
 log_cli_level = "DEBUG"
-addopts = "-m 'not asyncio'"
diff --git a/tests/integration/test_control_tasks.py b/tests/integration/test_control_tasks.py
index 3e6e362..490cb92 100644
--- a/tests/integration/test_control_tasks.py
+++ b/tests/integration/test_control_tasks.py
@@ -20,6 +20,9 @@ BASIC_CONFIG = {
             "sync_connection_factory": "dispatcherd.brokers.pg_notify.connection_saver",
             "default_publish_channel": "test_channel"
         }
+    },
+    "service": {
+        "process_manager_cls": "ProcessManager"
     }
 }
 ```

And then you're able to run all the tests together with:

```
pytest tests/
```

I don't know what kind of black magic allowed that to work, but I can't consider it _better_ than running separate tests. To me it seems really fast & loose with process management. I just want to disclose it. We _could_ run all the tests in a separate session, but at this point I'm no longer convinced that's a _good_ idea.

File: dispatcherd/service/process.py [hunk 2]
```diff
@@ -75,7 +78,7 @@ class ProcessManager:
 
     def __init__(self, settings: LazySettings = global_settings) -> None:
         self.ctx = multiprocessing.get_context(self.mp_context)
-        self.finished_queue: multiprocessing.Queue = self.ctx.Queue()
+        self._finished_queue: Optional[multiprocessing.Queue] = self.ctx.Queue()
```
chrismeyersfsu, line 81: The reason for adding `Optional` isn't clear to me. Can `self.ctx.Queue()` return `None` ?

dispatcherd/service/process.py
```diff
@@ -107,10 +118,20 @@ async def read_finished(self) -> dict[str, Union[str, int]]:
         message = await self.get_event_loop().run_in_executor(None, self.finished_queue.get)
         return message
 
+    def shutdown(self) -> None:
+        if self._finished_queue:
+            logger.debug('Closing finished queue')
```
chrismeyersfsu, line 123: Would be nice to know exactly which worker is emitting this log message.

AlanCoding: Oh... I guess that's the non-obvious thing. There is only one finished queue.

chrismeyersfsu (approved):

PR #130, opened 2025-03-18
Split task-delayer into new module, it is now a task "processor"

AlanCoding:
This stitches together 2 other refactors I already landed:

 - Introduce the generic notion of a task "processor" which can "consume" a task at various points with some implementation differences based on the order in the call stack.
   - The delayer runs early, and has the special ability of running an asyncio task. End result of tasks it consumes is a callback that passes it to the next stage of processing.
   - The blocker avoids running 2 tasks of the same method and call args based on specified `on_duplicate` behavior. End result of consumed tasks is a queue that is consumed in `drain_queue`, or discarded
   - The queuer has access to pool workers, and it places or consumes a task based on current capacity in the pool End result of consumed tasks is, again, a queue consumed by `drain_queue`. End result of non-consumed tasks is placement to worker.
 - Introduce a generic next-run-runner

You can see the whole deal here is that the 1st one, the task-delayer, is both a "processor" and (we'll call it) a "runner". This is not problematic philosophically or otherwise. It's just that I was working on these 2 constructs in parallel and now they are merged so I can do this.

### DIFF
--- a/dispatcherd/service/delayer.py
+++ b/dispatcherd/service/delayer.py
@@ -1,0 +1,84 @@
+import asyncio
+import logging
+import time
+from typing import Any, Callable, Coroutine, Iterator, Optional, cast
+
+from ..protocols import DelayCapsule as DelayCapsuleProtocol
+from ..protocols import Delayer as DelayerProtocol
+from .next_wakeup_runner import HasWakeup, NextWakeupRunner
+
+logger = logging.getLogger(__name__)
+
+
+class DelayCapsule(HasWakeup, DelayCapsuleProtocol):
+    """When a task has a delay, this tracks the delay, as in a time capsule"""
+
+    def __init__(self, delay: float, message: dict) -> None:
+        self.has_ran: bool = False
+        self.received_at = time.monotonic()
+        self.delay = delay
+        self.message = message
+
+    def next_wakeup(self) -> Optional[float]:
+        if self.has_ran is True:
+            return None
+        return self.received_at + self.delay
+
+
+class Delayer(NextWakeupRunner, DelayerProtocol):
+    def __init__(
+        self,
+        process_message_now: Callable[[dict[Any, Any]], Coroutine[Any, Any, tuple[str | None, str | None]]],
+        exit_event: Optional[asyncio.Event] = None,
+    ) -> None:
+        self.delayed_messages: set[DelayCapsuleProtocol] = set()
+        self.shutting_down = False
+        self.process_message_now = process_message_now
+        super().__init__(
+            wakeup_objects=cast(set[HasWakeup], self.delayed_messages),
+            process_object=self.run_delayed_capsule,  # type: ignore[arg-type] # takes capsules, not HasWakeups, which is more specific
+            name='delayed_task_runner',
+            exit_event=exit_event,
+        )
+
+    def __iter__(self) -> Iterator[DelayCapsuleProtocol]:
+        return iter(self.delayed_messages)
+
+    async def shutdown(self) -> None:
+        self.shutting_down = True
+        await super().shutdown()
+        for capsule in self.delayed_messages:
+            logger.warning(f'Abandoning delayed task (due to shutdown) to run in {capsule.delay}, message={capsule.message}')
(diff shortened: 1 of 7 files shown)


dispatcher/service/control_tasks.py
```diff
@@ -41,13 +41,13 @@ async def _find_tasks(dispatcher, cancel: bool = False, **data) -> dict[str, dic
...
             ret[f'queued-{i}'] = message
-    for i, capsule in enumerate(dispatcher.delayed_messages.copy()):
+    for i, capsule in enumerate(list(dispatcher.delayer)):
         if task_filter_match(capsule.message, data):
             if cancel:
                 uuid = capsule.message.get('uuid', '<unknown>')
                 logger.warning(f'Canceling delayed task (uuid={uuid})')
                 capsule.has_ran = True  # make sure we do not run by accident
-                dispatcher.delayed_messages.remove(capsule)
+                dispatcher.delayer.delayed_messages.remove(capsule)
```
AlanCoding, line 50: I do more of this in https://github.com/ansible/dispatcherd/pull/121

AlanCoding:
And ultimately, yes, I envision all of these being added to the protocols. Let me be specific - I envision delayer, blocker, and queuer being added to the protocols.

Defining a generic "processor" in protocols does not seem helpful. This is subtle so let me explain why.

A "processor" is not useful to an app using the dispatcher. It is only useful in-as-much as it is called in the task processing stack. That is true for delayer/blocker/queuer. Even if a delayer doesn't delay, or a blocker doesn't block, it still has a place where it is called in the core code base of main/pool classes. A user could _subclass_ any of these 3 and do something completely wild and boutique for their own needs.

dispatcherd/service/delayer.py
```diff
@@ -0,0 +1,84 @@
...
+        await super().shutdown()
+        for capsule in self.delayed_messages:
+            logger.warning(f'Abandoning delayed task (due to shutdown) to run in {capsule.delay}, message={capsule.message}')
+        self.delayed_messages = set()
+
+    async def create_delayed_task(self, delay: float, message: dict) -> None:
+        "Called as alternative to sending to worker now, send to worker later"
+        capsule = DelayCapsule(delay, message)
+        logger.info(f'Delaying {capsule.delay} s before running task: {capsule.message}')
+        self.delayed_messages.add(capsule)
```
pb82, line 58: Since we're adding to a Set here, is the default `__hash__` implementation sufficient? Or should we provide a custom one for `DelayCapsule`?

pb82 (approved):
LGTM, runs fine and tests pass. I think I understand the change: instead of having a `NextWakeupRunner` instance, we're using this as a base class now and `DelayCapsule` inherits from it. Simplifies the main loop.

Just a question about hashing behaviour (probably just my lack of Python experience). It doesn't look like you need to provide a custom __hash__ very often, at least for simple enough classes.

dispatcherd/service/delayer.py
AlanCoding: In terms of algorithms, what I'm counting on is that `capsule = DelayCapsule(delay, message)` creates a uniquely identifiable object. Even if `(delay, message)` are exactly repeated for 2 objects, I actually still want those to be treated as unique objects, and not get de-referenced by surprise.

This works as intended

```
>>> from dispatcherd.service.main import DelayCapsule
>>> 
>>> c1 = DelayCapsule(1, {})
>>> c2 = DelayCapsule(1, {})
>>> 
>>> s = set()
>>> s.add(c1)
>>> s.add(c2)
>>> s.add(c1)
>>> s
{<dispatcherd.service.main.DelayCapsule object at 0x7f4d5d391250>, <dispatcherd.service.main.DelayCapsule object at 0x7f4d5d3904d0>}
```

But that's not the exact reason I feel safe in these assumptions. I take it for granted that vanilla object instantiation creates a new entry in memory (for that object) and corresponding to that has a unique hash. It used to be that the hash _was_ the memory address but python made this complicated somehow. Nonetheless, if an object has its own space in memory, there's a firm expectation that the natural hash function maps to that memory space, which is what my snippet above is leaning on.

PR #133, opened 2025-03-21
Detect broken connections and restart producer

pb82:
Fixes https://github.com/ansible/dispatcherd/issues/2


dispatcher/brokers/pg_notify.py
```diff
@@ -129,6 +163,41 @@ def get_unlisten_query(self) -> psycopg.sql.SQL:
         """Stops listening on all channels for current session, see pg_notify docs"""
         return psycopg.sql.SQL("UNLISTEN *;")
 
+    async def initiate_self_check(self) -> None:
+        if self.self_check_status == BrokerSelfCheckStatus.IN_PROGRESS:
+            # another self-check message is in flight
+            raise RuntimeError(f'failed to initiate broker self check on node {self.node_id}')
```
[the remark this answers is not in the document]
pb82: I've removed the node id requirement, it now uses a uuid per broker.

dispatcher/brokers/pg_notify.py
```diff
@@ -129,6 +161,37 @@ def get_unlisten_query(self) -> psycopg.sql.SQL:
...
+        except Exception as error:
+            raise RuntimeError(f'failed to send self check message: {error}')
+
+    def verify_self_check(self, message: dict[str, Any]) -> None:
+        """Verify a received self check message: check if it was sent from the same node and
+        is not outdated
+        """
+        now = time.time()
+        message_date = message['sent']
+        delta_seconds = now - message_date
```
AlanCoding, line 181: No problem sending the "time" in the message, but I would actually prefer to save the time on `self` when using it for comparison. This only requires that we only are running 1 self-check at a time, which I think is a safe assumption.

That would make it better to use `time.monotonic()` for, but that's not a meaningful global time stamp. If you are sending something in a message `time.time()` is better as a true epoch time, but `time.monotonic()` is less vulnerable to jumps due to time sync.

dispatcher/brokers/pg_notify.py
```diff
@@ -129,6 +161,37 @@ def get_unlisten_query(self) -> psycopg.sql.SQL:
...
+    async def initiate_self_check(self) -> None:
+        if self.self_check_status == BrokerSelfCheckStatus.IN_PROGRESS:
+            # another self-check message is in flight
+            raise RuntimeError(f'failed to initiate self check for broker {self.broker_id}')
+
+        try:
+            await self.apublish_message(self.self_check_channel, json.dumps({'self_check': True, 'sent': time.time(), 'task': f'lambda: "{self.broker_id}"'}))
+            self.self_check_status = BrokerSelfCheckStatus.IN_PROGRESS
+        except Exception as error:
+            raise RuntimeError(f'failed to send self check message: {error}')
```
AlanCoding, line 173: Mind if we get rid of this try-except block?

If this raises an exception, I would prefer to treat that as a failure of the check outright.

pb82: Agree, this block actually doesn't do anything, it just catches and re-raises.

dispatcher/brokers/pg_notify.py
```diff
@@ -63,6 +70,18 @@ def __init__(
         if not (config or sync_connection_factory or sync_connection):
             raise RuntimeError('Must specify either config or sync_connection_factory')
 
+        if not max_connection_idle_seconds:
+            raise RuntimeError('Must specify max_connection_idle_seconds')
```
AlanCoding, line 74: Let me formally describe a change I would like to make, just to add it to the list -

If `max_connection_idle_seconds` is None, I would like for this to signal "self-checks are disabled" as opposed to raising an exception. It's already `Optional`. I'm good with your default, but a user should be able to set to `None` and turn self-checks off.

dispatcher/protocols.py
```diff
@@ -40,6 +48,10 @@ def close(self):
         """Close the sychronous connection"""
         ...
 
+    def verify_self_check(self, message: dict[str, Any]) -> None:
+        """Verify a received self check message"""
+        ...
```
AlanCoding, line 53: :+1:  Nothing to add here, just want to see this is nice to see.

Needs to be added to the protocol because it is necessary to be used by `DispatcherMain`, making it a public method. I was worried that these would be confusing and I was going too far doubling-down on these (often necessitated as I went further down mypy checks).

AlanCoding:
I pushed a commit after testing it separately. A couple of things I'll describe here:

 - Some minor code moves and asserts are done to satisfy type hints
 - Ordering in `verify_self_check` is reversed, because I expected a bug where a peer node sends a message before the local node has ever sent its first health check message
 - Cleared out some stuff in `aclose`, because this addresses novel concerns introduced by recycling
 - Moved `wait_for_any` to the asyncio utils, because those are only used by the service
 - I had to move some steps in `DispatcherMain.main` into other methods for the test
 - Updated config schema
 - Added test that drops connection based on `connection_name`
 - Increased psycopg min version to 3.2.4 because I discovered the fix https://github.com/psycopg/psycopg/pull/975 is needed - thanks @dvarrazzo, you don't know me, but your work has been critical for what we've been doing, and you have my humble appreciation

AlanCoding (approved):

dvarrazzo:
> @dvarrazzo, you don't know me, but your work has been critical for what we've been doing, and you have my humble appreciation

I have been an ansible user in the past and I know I will be again in the future. If I have been helpful I can only consider it a honour.

PR #147, opened 2025-04-07
Fix demo build by installing git

AlanCoding:
Dynamic versioning added in https://github.com/ansible/dispatcherd/pull/135 requires git, or else the `pip install` command will error. This adds the git install so that the demo docker-compose can build again.

### DIFF
--- a/tools/demo/Dockerfile
+++ b/tools/demo/Dockerfile
@@ -8,6 +8,9 @@ RUN python3 -m venv /dispatcherd/venv
 
 RUN echo $pwd
 
+# Git is necessary because of setuptools-scm
+RUN apk add --no-cache git
+
 COPY . ./src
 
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo (approved):

AlanCoding:
> environment where `.git` is not available

The error has a lot of noise in it, but I focused on these lines from the output:

```
1.873   WARNING setuptools_scm.run_cmd command git missing: [Errno 2] No such file or directory: 'git'
1.874   ERROR setuptools_scm._get_version_impl command git not found while parsing the scm, using fallbacks
```

I think this is telling me that `git` is not available, as opposed to `.git/`. Checking back over the Dockerfile, I actually copy the git history of the repo. This might still error if tags aren't fetched or something.

Alex-Izquierdo:
@AlanCoding Indeed, it seems a lack of git binary. I'm just thinking in cases where the .git is excluded which is an usual practice (not our case though)

Issue #149, opened 2025-04-07
Perpetual mutual-blocking when more than 1 task is blocked


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcherd/issues) for duplicates.
- [x] I understand that dispatcherd is open source software provided for free and that I might not receive a timely response.

### Bug Summary

There is a bug draining the queue of a series of blocked tasks. Here:

https://github.com/ansible/dispatcherd/blob/f486d8e57e544ff9426a3c516574d2cbf2855b4a/dispatcherd/service/blocker.py#L84

Introduced in https://github.com/ansible/dispatcherd/pull/111

The issue is that `self.already_queued(message)` may always be `True` if 2 or more of a task is submitted. Reproduced in demo at:

https://github.com/ansible/dispatcherd/blob/f486d8e57e544ff9426a3c516574d2cbf2855b4a/run_demo.py#L134-L136

### Environment

n/a

### Steps to reproduce

Run the demo

### Actual results

The queue of blocked tasks clears some time (1-2 min for this) after the demo finishes.

### Expected results

The backlog of tasks is never cleared.

### Additional information

I believe the previous code had this right, although more complicated than what was needed.

```python
    def get_blocking_action(self, message: dict) -> str:
        on_duplicate = message.get('on_duplicate', DuplicateBehavior.parallel.value)

        if on_duplicate == DuplicateBehavior.serial.value:
            if self.already_running(message):
                return MessageAction.queue.value

        elif on_duplicate == DuplicateBehavior.discard.value:
            if self.already_running(message) or self.already_queued(message):
                return MessageAction.discard.value

        elif on_duplicate == DuplicateBehavior.queue_one.value:
            if self.already_queued(message):
                return MessageAction.discard.value
            elif self.already_running(message):
                return MessageAction.queue.value

        elif on_duplicate != DuplicateBehavior.parallel.value:
            logger.warning(f'Got unexpected on_duplicate value {on_duplicate}')

        return MessageAction.run.value
```

Specifically, the `if on_duplicate == DuplicateBehavior.serial.value:` case (here, from demo reproducer), checks if the task is already running. If not, then it gives the default `MessageAction.run.value`, meaning, "run the task".

[...]

PR #150, opened 2025-04-07
Fix queue draining logic

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/149

### DIFF
--- a/dispatcherd/service/blocker.py
+++ b/dispatcherd/service/blocker.py
@@ -28,7 +28,7 @@ class Blocker(BlockerProtocol):
         return False
 
     def already_running(self, message: dict) -> bool:
-        return self._duplicate_in_list(message, self.queuer.running_tasks())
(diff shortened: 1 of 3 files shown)


AlanCoding:
tested with the demo and it's looking good.

art-tapin (approved):
I've tested the PR by running several scenarios (basic serial execution, high volume serial tasks, mixed behaviors, cancellation, and submissions during drain).

The results look solid - most importantly, the high volume serial test with 10 identical tasks running with `on_duplicate="serial"` now processes all tasks sequentially without any deadlock. Before your fix, these would have been stuck in a perpetual blocking situation.

I particularly like how you simplified the logic by changing the condition in `pop_unblocked_messages()` to check against the already-unblocked list instead of the blocked queue - very clean

The cancellation and mixed behavior scenarios taht I tested also pass perfectly :+1: 

LGTM, merging this in! 

:1st_place_medal:

Issue #140, opened 2025-03-24
Pass a shared object (in factories) for global `exit_event` and locks


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

I have a lot of this sprinkled around the code:

https://github.com/ansible/dispatcherd/blob/299d7ea37ac37a37b8aae906fc2f3a2e773f54b2/dispatcherd/service/pool.py#L210

The library effectively shifted away from this pattern as `exit_event` started being used to exit the program. It was needed to use this event as it got sent to `add_done_callback`.

But this is problematic, because it means that any line that starts a new task needs to have access to the global `DispatcherMain` object, and it shouldn't have that role.

We should create a new object... don't know what to call it, like `SharedObjects`. This would contain the lock for forking and connecting, as well as the exit event.

Then we would no longer maintain `self.shutting_down` in classes. Instead, we would reference `exit_event.is_set()` on the `SharedObjects` object, that nearly all classes would have passed to their `__init__`, probably.

### Steps to reproduce

refactoring

### Sugested feature result

refactoring

PR #153, opened 2025-04-11
Use a shared set of asyncio objects for code flow control

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/140

The code might not always look cleaner as a result of this change, but more importantly, the internal state of the program is simplified. There are no longer multiple `shutting_down` flags. There's just 1 event, and anything can either check it or wait for it. Those other flags were subservient to this event (which already existed anyway). The ownership of that `exit_event` is changed from `DispatcherMain` to "anyone can pull the fire alarm".

### DIFF
--- a/dispatcherd/config.py
+++ b/dispatcherd/config.py
@@ -76,3 +76,8 @@ def temporary_settings(config):
         yield settings
     finally:
         settings._wrapped = prior_settings
+
+
+def is_setup(for_settings: LazySettings = settings) -> bool:
+    """Tells whether dispatcherd has been configured"""
+    return for_settings._wrapped is None
--- a/dispatcherd/factories.py
+++ b/dispatcherd/factories.py
@@ -9,6 +9,7 @@ from .config import settings as global_settings
 from .control import Control
 from .protocols import Broker, Producer
 from .service import process
+from .service.asyncio_tasks import SharedAsyncObjects
 from .service.main import DispatcherMain
 from .service.pool import WorkerPool
 
@@ -29,25 +30,33 @@ def process_manager_from_settings(settings: LazySettings = global_settings) -> p
     return process_manager_cls(**kwargs)
 
 
-def pool_from_settings(settings: LazySettings = global_settings) -> WorkerPool:
+def pool_from_settings(shared, settings: LazySettings = global_settings) -> WorkerPool:
     kwargs = settings.service.get('pool_kwargs', {}).copy()
     kwargs['process_manager'] = process_manager_from_settings(settings=settings)
+    kwargs['shared'] = shared
     return WorkerPool(**kwargs)
 
 
-def brokers_from_settings(settings: LazySettings = global_settings) -> Iterable[Broker]:
-    return [get_broker(broker_name, broker_kwargs) for broker_name, broker_kwargs in settings.brokers.items()]
+def brokers_from_settings(settings: LazySettings = global_settings, shared: Optional[SharedAsyncObjects] = None) -> Iterable[Broker]:
+    brokers = []
+    for broker_name, broker_kwargs in settings.brokers.items():
+        kwargs = broker_kwargs.copy()
+        if shared:
+            kwargs['shared'] = shared
+        brokers.append(get_broker(broker_name, kwargs))
+    return brokers
 
 
-def producers_from_settings(settings: LazySettings = global_settings) -> Iterable[Producer]:
+def producers_from_settings(shared: SharedAsyncObjects, settings: LazySettings = global_settings) -> Iterable[Producer]:
     producer_objects = []
     for broker in brokers_from_settings(settings=settings):
-        producer = producers.BrokeredProducer(broker=broker)
+        producer = producers.BrokeredProducer(shared=shared, broker=broker)
         producer_objects.append(producer)
 
     for producer_cls, producer_kwargs in settings.producers.items():
         if producer_kwargs is None:
             producer_kwargs = {}
+        producer_kwargs['shared'] = shared
         producer_objects.append(getattr(producers, producer_cls)(**producer_kwargs))
 
     return producer_objects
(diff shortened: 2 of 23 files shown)


pb82:
@AlanCoding the sharing here relies on only one instance of `SharedAsyncObjects` existing and then passing it down through the chain right? I wonder if it would make sense to use something like `functools.cache` (or a decorator, or a singleton) to enforce a single instance?

Then we wouldn't need to pass the shared instance everywhere and could just refer to the single instance wherever we need it.

AlanCoding:
but I missed 1 case of `shutting_down` and will push a commit to get rid of that.

pb82:
> It just seems like that would go against best practice.
> 
> What if you ran 2 dispatcherd services in the same event loop? That's why the factories take `settings` as an input, and why I specifically do not use the _singleton settings_ object _without an override_. I'm not saying that 2 co-running dispatcherd services is practical for production use... but for tests? Could be.
> 
> It's also kind of my entire point to of having the factories module - to absorb any object construction complexity that comes about from this. That creates very fragmented objects with minimal cross-references. Sure this is annoying in some ways, but also a godsend in other ways.

Ok, understand. I wasn't thinking about the scenario of running two or more dispatcherd instances.

pb82:
@AlanCoding when I run dispatcher from this branch, this error pops up in the log from time to time:

```
ERROR:dispatcherd.service.pool:Unexpected error starting worker 13 subprocess, marking as error
  File "/home/peter/PycharmProjects/dispatcher/dispatcherd/service/pool.py", line 44, in start
    self.process.start()
    ~~~~~~~~~~~~~~~~~~^^
  File "/home/peter/PycharmProjects/dispatcher/dispatcherd/service/process.py", line 30, in start
    self._process.start()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.13/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
                  ~~~~~~~~~~~^^^^^^
  File "/usr/lib64/python3.13/multiprocessing/context.py", line 301, in _Popen
    return Popen(process_obj)
  File "/usr/lib64/python3.13/multiprocessing/popen_forkserver.py", line 35, in __init__
    super().__init__(process_obj)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/multiprocessing/popen_fork.py", line 20, in __init__
    self._launch(process_obj)
    ~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/multiprocessing/popen_forkserver.py", line 47, in _launch
    reduction.dump(process_obj, buf)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
TypeError: cannot pickle '_asyncio.Future' object
```

It doesn't happen on the main branch, so I suspect it has to do with this change.

AlanCoding:
Oh I saw that in some other case. I think it's because another PR changed the arg order to `work_loop` and tests or something else got the args jumbled. Will look.

AlanCoding:
A lot changed with the rebase, but after doing that, I can reproduce any version of the pickling error. It is concerning that I have some memory of it.

Another way this might have happened is if a "future" object made its way into settings, coming out via `settings.serialize()`. It's easy to see how that might happen due to bugs in the config processing. It might be good to add some validation to that method to ensure JSON type content. But... it doesn't seem to be happening for me, now.

AlanCoding:
How are you starting things? Is this from running the service manually? Same demo config?? User the docker-compose? Running some tests? If you're not on the same OS, it's very possible that cross-platform issues can come up.

AlanCoding:
If you want another debugging approach, you could put a print statement before this:

https://github.com/ansible/dispatcherd/blob/7b529bf3e40a56671aa7f073d45ca40ba7628008/dispatcherd/service/process.py#L27

And print `(args, kwargs)`. Then look, is anything in there something "bad" like a coroutine. It should contain very static data, like strings, ints, and containers of those.

pb82:
I'm starting dispatcher with this script (using the default config that's in the repo):

```
#!/usr/bin/env python3

import logging
import sys

from dispatcherd import run_service
from dispatcherd.factories import get_publisher_from_settings
from dispatcherd.config import setup

# Setup the global config from the settings file shared with the service
setup(file_path='dispatcher.yml')

broker = get_publisher_from_settings()

def main():
    run_service()

if __name__ == "__main__":
    logging.basicConfig(level='DEBUG', stream=sys.stdout)
    main()
```

and this is the output of the print statement (right before the error occurs):

```
() {'worker_id': 2, 'settings': {'version': 2, 'brokers': {'pg_notify': {'config': {'conninfo': 'dbname=dispatch_db user=dispatch password=dispatching host=localhost port=55777 application_name=dispatcher_demo_run'}, 'sync_connection_factory': 'dispatcherd.brokers.pg_notify.connection_saver', 'channels': ['test_channel', 'test_channel2', 'test_channel3'], 'default_publish_channel': 'test_channel', 'max_connection_idle_seconds': 5, 'max_self_check_message_age_seconds': 2}, 'socket': {'socket_path': 'demo_dispatcher.sock'}}, 'producers': {'ScheduledProducer': {'task_schedule': {'lambda: __import__("time").sleep(1)': {'schedule': 3}, 'lambda: __import__("time").sleep(2)': {'schedule': 3}}, 'shared': <dispatcherd.service.asyncio_tasks.SharedAsyncObjects object at 0x7f7be14701a0>}, 'OnStartProducer': {'task_list': {'lambda: print("This task runs on startup")': {}}, 'shared': <dispatcherd.service.asyncio_tasks.SharedAsyncObjects object at 0x7f7be14701a0>}, 'ControlProducer': None}, 'service': {'pool_kwargs': {'min_workers': 2, 'max_workers': 12, 'scaledown_wait': 15}, 'main_kwargs': {'node_id': 'demo-server-a', 'metrics': <dispatcherd.service.metrics.DispatcherMetricsServer object at 0x7f7be0558440>}, 'metrics_kwargs': {'log_level': 'debug'}}, 'publish': {'default_control_broker': 'socket', 'default_broker': 'pg_notify'}}, 'finished_queue': <multiprocessing.queues.Queue object at 0x7f7be14717f0>, 'message_queue': <multiprocessing.queues.Queue object at 0x7f7be057bbb0>}
```

AlanCoding:
@pb82 fantastic find! There was 1 single case where I missed a `.copy()` for a dictionary in the settings.

It looks like you managed to hit this by making your `broker` early. Most likely, I was creating the `WorkerPool` before the brokers, so the bug didn't manifest.

In my patch, I did what I mentioned earlier, which is to turn the settings into a JSON string, which makes it stricter than any test I could have written. With the JSON conversion, a large fraction of the tests failed until I fixed the `.copy()` bug. I think this should do good for regression-prevention.

pb82 (approved):

PR #154, opened 2025-04-12
Add a metrics server that can be optionally enabled

AlanCoding:
Replaces https://github.com/ansible/dispatcherd/pull/116

This takes a pivot from what I initially suggested, which was to have AWX start the thread.

That was a terrible idea, because nobody anywhere should be starting any threads here. Our hangup was that the library `prometheus-client` didn't obviously implement an asyncio version of their server. Oh, but they did! It just needs an intermediary of uvicorn.

After learning about how we can use this full-asyncio solution, I'm very comfortable with implementing the entire thing natively in dispatcherd. This PR is intended to offer the meat of the integration work for that - putting the metrics kwargs into the dispatcherd config system, adding that tasks to our tasks.

By default this will not be enabled, which is what you get if you don't include `metrics_kwargs`. This seems for the best, because you really shouldn't be accepting whatever default port it runs on in AWX or eda-server integration. Also, the dependencies are added as an _extra_ because I expect we'll have a bunch of dependency issues once this gets used downstream.

### DIFF
--- a/dispatcher.yml
+++ b/dispatcher.yml
@@ -8,6 +8,8 @@ service:
     scaledown_wait: 15  # seconds
   main_kwargs:
     node_id: demo-server-a
+  metrics_kwargs:
+    log_level: debug
 brokers:
   pg_notify:
     config:
--- a/dispatcherd/factories.py
+++ b/dispatcherd/factories.py
@@ -62,6 +62,13 @@ def from_settings(settings: LazySettings = global_settings) -> DispatcherMain:
     producers = producers_from_settings(settings=settings)
     pool = pool_from_settings(settings=settings)
     extra_kwargs = settings.service.get('main_kwargs', {})
+
+    metrics_kwargs = settings.service.get('metrics_kwargs')
+    if metrics_kwargs:
+        from .service.metrics import DispatcherMetricsServer
+
+        extra_kwargs['metrics'] = DispatcherMetricsServer(**metrics_kwargs)
+
     return DispatcherMain(producers, pool, **extra_kwargs)
 
 
@@ -105,7 +112,7 @@ def get_control_from_settings(publish_broker: Optional[str] = None, settings: La
 
 # ---- Schema generation ----
 
-SERIALIZED_TYPES = (int, str, dict, type(None), tuple, list, float)
+SERIALIZED_TYPES = (int, str, dict, type(None), tuple, list, float, bool)
 
 
 def is_valid_annotation(annotation):
@@ -134,6 +141,9 @@ def generate_settings_schema(settings: LazySettings = global_settings) -> dict:
 
     ret['service']['pool_kwargs'] = schema_for_cls(WorkerPool)
     ret['service']['main_kwargs'] = schema_for_cls(DispatcherMain)
+    from .service.metrics import DispatcherMetricsServer
+
+    ret['service']['metrics_kwargs'] = schema_for_cls(DispatcherMetricsServer)
     ret['service']['process_manager_kwargs'] = {}
     pm_classes = (process.ProcessManager, process.ForkServerManager)
     for pm_cls in pm_classes:
--- a/dispatcherd/service/metrics.py
+++ b/dispatcherd/service/metrics.py
@@ -1,0 +1,74 @@
+import logging
+from typing import Any, Generator
+
+# Metrics library
+from prometheus_client import CollectorRegistry, make_asgi_app
+
+# For production of the metrics
+from prometheus_client.core import CounterMetricFamily
+from prometheus_client.metrics_core import Metric
+from prometheus_client.registry import Collector
+
+# General ASGI python web server, interfaces with prometheus-client lib by the ASGI standard
+from uvicorn.config import Config
+from uvicorn.server import Server
+
+from ..protocols import DispatcherMain
+
+logger = logging.getLogger(__name__)
+
+
+def metrics_data(dispatcher: DispatcherMain) -> Generator[Metric, Any, Any]:
+    """
+    Called each time metrics are gathered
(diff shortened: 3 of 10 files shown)


AlanCoding:
Metrics to add from comments in prior PR:

 - Dispatcher pool scale-up events, total up to the present moment
 - Dispatcher active tasks at the moment of collection

dispatcherd/factories.py
```diff
@@ -10,6 +10,7 @@
 from .protocols import Broker, Producer
 from .service import process
 from .service.main import DispatcherMain
+from .service.metrics import DispatcherMetricsServer
```
AlanCoding, line 13: No, can't do this because of case that extras aren't installed. Will move.

pyproject.toml
```diff
@@ -64,6 +64,7 @@ line_length = 160
 
 [project.optional-dependencies]
 pg_notify = ["psycopg[binary]>=3.2.4"]
+metrics = ["uvicorn[standard]", "prometheus-client"]
```
pb82, line 67: should that really be optional?

pb82 (approved):
The endpoint returns the following metrics:

```
# HELP dispatcher_messages_received_total Number of messages received by dispatchermain
# TYPE dispatcher_messages_received_total counter
dispatcher_messages_received_total 3.0
# HELP dispatcher_control_messages_count_total Number of control messages received.
# TYPE dispatcher_control_messages_count_total counter
dispatcher_control_messages_count_total 0.0
# HELP dispatcher_worker_count_total Number of workers running.
# TYPE dispatcher_worker_count_total counter
dispatcher_worker_count_total 2.0
```

Looks good, there are probably other interesting metrics that we could add later:

* how many times each broker was recycled
* how long it takes to run tasks (as a histogram with buckets)

dispatcher.yml
```diff
@@ -8,6 +8,8 @@ service:
     scaledown_wait: 15  # seconds
   main_kwargs:
     node_id: demo-server-a
+  metrics_kwargs:
+    log_level: debug
```
chrismeyersfsu, line 12: wondering what this does .. I will continue reading

dispatcherd/service/metrics.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, dispatcher: DispatcherMain) -> None:
+        self.dispatcher = dispatcher
+
+    def collect(self) -> Generator[Metric, Any, Any]:
+        for m in metrics_data(self.dispatcher):
+            yield m
+
+
+class DispatcherMetricsServer:
+    def __init__(self, port: int = 8070, log_level: str = 'info', host: str = "localhost") -> None:
```
chrismeyersfsu, line 53: Sweet, you showed the `log_level` being passed through the config so the `port` and `host` can also be configured via the dispatcher yaml :+1:

tests/integration/test_metrics_use.py
```diff
@@ -0,0 +1,53 @@
...
+
+TEST_METRICS_PORT = 18080
+
+
+@pytest.fixture(scope='session')
+def metrics_config():
+    return {
+        "version": 2,
+        "brokers": {},
+        "service": {"main_kwargs": {"node_id": "metrics-test-server"}, "metrics_kwargs": {"log_level": "debug", "port": TEST_METRICS_PORT}},
```
chrismeyersfsu, line 22: Oh, and here is the port being passed through :+1:

chrismeyersfsu (approved):

dispatcherd/service/metrics.py
```diff
@@ -0,0 +1,74 @@
...
+from prometheus_client import CollectorRegistry, make_asgi_app
+
+# For production of the metrics
+from prometheus_client.core import CounterMetricFamily
+from prometheus_client.metrics_core import Metric
+from prometheus_client.registry import Collector
+
+# General ASGI python web server, interfaces with prometheus-client lib by the ASGI standard
+from uvicorn.config import Config
+from uvicorn.server import Server
```
chrismeyersfsu, line 14: Looks like you can maybe use a library that comes with python ..

```
from wsgiref.simple_server import make_server
```

```
from prometheus_client import make_wsgi_app, Counter
from wsgiref.simple_server import make_server

# Define a Prometheus counter metric
request_counter = Counter("http_requests_total", "Total number of HTTP requests.")

def simple_app(environ, start_response):
    """A simple WSGI application."""
    status = '200 OK'
    headers = [('Content-type', 'text/plain')]
    start_response(status, headers)
    request_counter.inc()
    return [b"Hello, World!\n"]

# Create the Prometheus WSGI application
metrics_app = make_wsgi_app()

def application(environ, start_response):
    """A combined WSGI application, routing /metrics to Prometheus."""
    if environ['PATH_INFO'] == '/metrics':
        return metrics_app(environ, start_response)
    else:
        return simple_app(environ, start_response)

if __name__ == '__main__':
    httpd = make_server('', 8000, application)
    print("Serving on port 8000...")
    httpd.serve_forever()
```

schema.json
```diff
@@ -35,6 +35,11 @@
     "main_kwargs": {
       "node_id": "typing.Optional[str]"
     },
+    "metrics_kwargs": {
```
chrismeyersfsu, line 38: I missed this. This (should) make what I've been looking for obvious.

pyproject.toml
AlanCoding: A metrics server via prometheus-client is kind of a boutique thing.

Maybe I'm being a little too die-hard about this, but imagine that someone specifies a metrics class in the config that was something entirely different, using an entirely different stack. As long as that class conformed to the protocol, it could run and serve something entirely different on some port.

AlanCoding:
For adding more metrics

https://github.com/ansible/dispatcherd/issues/158

Issue #143, opened 2025-03-24
Another utility to publish a task


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

Further building on https://github.com/ansible/dispatcherd/issues/142

If you look at how rq-worker submits tasks, they'll do this:

https://python-rq.org/docs/jobs/

```python
job = q.enqueue(count_words_at_url, 'http://nvie.com')
```

Whereas we're still using these old attached `.delay` and `.apply_async` from the Celery days.

We don't have to get rid of those (but could configure it to not attach them maybe), but this asks that we add a new interface for submitting tasks. Spitball:

```python
from dispatcherd.publisher import publish_task
from dispatcherd.processors.delayer import Delayer

publish_task(
  count_words_at_url,
  args=('http://foo.invalid',),
  delayer=Delayer.publish_options(delay=4)
)
```

The reason for using this construction is so that we can type-hint options in a plug-gable form.

Note, in my head, we _may still require_ that the `@task()` decorator be applied to the task submitted. This allows for validation, and restricting running tasks to only those that are registered. Or you could just run whatever you want if you don't want that feature.

### Steps to reproduce

enhancement

### Sugested feature result

enhancement

Alex-Izquierdo:
@AlanCoding Based on your example, `publish_task` function should not require the channel argument?

AlanCoding:
It would. Good point. I didn't declare that, it would take many of the existing kwargs to `@task`

https://github.com/ansible/dispatcherd/blob/ed4fde53131c7b996e21c88c107ba513f4277742/dispatcherd/publish.py#L38-L45

That's because right now many of these can already be overridden by the current method, meaning `.apply_async(uuid='1234')` (which co-exists along with `args` type stuff), and this spitball of `publish_task` would be a direct replacement for `apply_async`.

Out of those, only
 - `timeout` and 
 - `on_duplicate`

can move to the new system.

The
 - `uuid`
 - `queue`, and 
 - `bind`, plus
 -  `decorate` (speculative, new)

[...]

Issue #142, opened 2025-03-24
Add a "processors" section of the config


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

Companion to https://github.com/ansible/dispatcherd/issues/138

I see the config specifying these options via:

```yaml
version: 2
processors:
  validator_kwargs:
    allow_lambda_functions: false
```

Right now, the code is under `dispatcherd.service`.

Why?

Because while the processors are only used by the service now, this would make sense to attach to the message on publication in some cases (although clearly not in this case).

As another example, the "delay" kwarg to `@task` should be "owned" by the processor. So the processor should arguably be usable by the publisher at some point. That makes these shared entities, more similar to brokers than the other "service" objects.

### Steps to reproduce

refactoring

### Sugested feature result

refactoring

Issue #138, opened 2025-03-24
Validate message body


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

We should add a validator in the main process that may _optionally_ validate things, hopefully obvious things, like that the message has a "task" specified in it.

You can keep going and going beyond that.
1. Disallow running `lambda` functions, which are only used for testing - this is relatively high priority
2. Disallow running methods not currently seen in the registry
3. Disallow tasks that have unexpected keys in the body

Code-wise, I believe this should be implemented as a "processor".

### Steps to reproduce

I manually experimented with `dispatcherctl` some, and I found out what happens if you send:

```json
{}
```

As the message.

The main program just passes it along to a worker. This is much better than the other option, which is crashing. But still not great.

### Sugested feature result

message validation options.

PR #163, opened 2025-04-29
Add a static `submit_task` method to avoid attached methods

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/143

It also works in the direction of https://github.com/ansible/dispatcherd/issues/142 and https://github.com/ansible/dispatcherd/issues/138 but does not do those in this patch.

AAP-43947


AlanCoding:
As a changelog matter - I plan to rename `parts` to something better. I'm still not sure what.

dispatcherd/registry.py
```diff
@@ -69,37 +89,58 @@ def get_async_body(
...
+
+        This submission does not provide confirmation.
+        If you need confirmation from the service, look into the "run" control command.
+        """
 
-        if callable(queue):
-            queue = queue()
+        resolved_queue: str
+        if queue is None:
+            resolved_queue = self.queue
```
Alex-Izquierdo, line 129: Since `self.queue` has also `None` as default and comes from the caller also with `None` as default, should we check it before the assignment or it is safe? I can not find where we ensure the value of the queue.

Alex-Izquierdo (approved):
LGTM. 

I feel all the processors parameters deserves its own documentation, I would have questions like: Can I provide my own processor? what are the available processors and how to use them? and so on...

dispatcherd/registry.py
AlanCoding: I wanted to be able to set the default in the config, and the most logical place was in the broker, so the ultimate enforcement has to be there:

https://github.com/ansible/dispatcherd/blob/dc16ffa6b6e8c097f48dde62f521c68982e2477d/dispatcherd/brokers/pg_notify.py#L126-L132

I'll leave a comment saying this. Might have to put that comment in 2 or 3 places.

dispatcherd/processors/params.py
```diff
@@ -0,0 +1,21 @@
...
+
+@dataclass(kw_only=True)
+class ProcessorParams:
+    """Data structure for task options, specific to a particular processor in the dispatcherd service"""
+
+    def to_dict(self) -> dict[str, Any]:
+        return asdict(self)
+
+    @classmethod
+    def from_message(cls: Type[T], message: dict[str, Any]) -> T:
```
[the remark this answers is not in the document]
AlanCoding: good tip, pushed a commit for this.

PR #164, opened 2025-04-30
chore: ensure httpx as dev dependency

Alex-Izquierdo:
Ensure httpx is a dev dependency and make requirements_dev as single source of truth for dependencies.


AlanCoding (approved):
Yeah, I had separately arrived at the same thing some other time, but didn't have it in.

PR #166, opened 2025-05-06
feat: set max_workers by default based on cpu count

Alex-Izquierdo:
In the most of the cases we want to set the max number of workers equal to the cpu cores available. This can be set by default by the system if the configuration does not provide a value, allowing a more concise configuration. 

It also removes the default at the instantiation level, unaligned with previous existing default value in the class.


AlanCoding (approved):
the default in the config was just an oversight, something forgotten about

PR #169, opened 2025-05-15
Add decorate kwarg to task decorator itself

AlanCoding:
This seems to be something that I outright forgot. This use is literally documented.

https://github.com/ansible/dispatcherd/blob/main/docs/submit_task.md

Discovered when trying to use it from AWX.


Alex-Izquierdo (approved):
LGTM, but it suggests there is a missing test somewhere

PR #171, opened 2025-05-15
Make queuing message a lower log level

AlanCoding:
Since I've been going through AWX logs a lot, I have been looking at a lot of log messages related to queuing tasks. But ordinarily, this is a non-story. If we have a very very high queue count, then it could be important.

So this approach uses a gradually increasing log level as queue pressure increases. I mainly just tested with the demo.

```
DEBUG:dispatcherd.service.pool:Dispatching task (uuid=internal-2039) to worker (id=6)
DEBUG:root:Queueing task (uuid=internal-2040), due to lack of capacity, queued_ct=0
INFO:root:Queueing task (uuid=internal-2041), due to lack of capacity, queued_ct=1
WARNING:root:Queueing task (uuid=internal-2042), due to lack of capacity, queued_ct=2
```

### DIFF
--- a/dispatcherd/processors/queuer.py
+++ b/dispatcherd/processors/queuer.py
@@ -9,6 +9,9 @@ from ..protocols import Queuer as QueuerProtocol
 logger = logging.getLogger(__name__)
 
 
+QUEUE_LVL_MAP = {0: logging.DEBUG, 1: logging.INFO}
+
+
 class Queuer(QueuerProtocol):
     @dataclass(kw_only=True)
     class Params(ProcessorParams):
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo (approved):
LGTM. 

One question. 
Maybe could it be dynamic? I wonder if it could depend on the number of subprocesses. It would not be the same 2 count with one worker than with 8. Even it could be a threshold overridable through environment configuration, to be able to say: in this case I expect to reach 5, for instance.

AlanCoding:
That makes less sense than what would be obvious. In the current implementation I'm throttling the subprocess spawning to 1-at-a-time. It would be possible to create 10 new workers (processes) simultaneously, and it might go faster than creating them in serial, but that consumes more serious OS resources and we never did this in AWX, so I am sacrificing some scale-up performance for conservatism.

Assuming you only have long-running tasks, and that you need to process a burst of messages only with scale-up, then it doesn't matter what the prior worker count is. You're going to scale a worker, place a task, scale a worker, place a task, etc.... Any queuing introduces _some_ delay, but because of the current implementation, the time for scale-up to get around to a task is linear with queue size.

If scale-up can't help to clear your queue, then eventually you'll get around to some more severe log messages due to bumping up against max workers.

Alex-Izquierdo:
Makes all the sense. Thanks for your clarification @AlanCoding

PR #179, opened 2025-05-27
Implement metrics server with python standard library

AlanCoding:


dispatcherd/service/metrics.py
```diff
@@ -49,26 +46,126 @@ def collect(self) -> Generator[Metric, Any, Any]:
...
+            status_line = "HTTP/1.1 404 Not Found"
+            content_type = "text/plain; charset=utf-8"
+
+        response_headers = (
+            f"{status_line}\r\n" f"Content-Type: {content_type}\r\n" f"Content-Length: {len(body.encode('utf-8'))}\r\n" "Connection: close\r\n\r\n"
+        )
+
+        response = response_headers + body
+
+        writer.write(response.encode('utf-8'))
```
chrismeyersfsu, line 110: I wonder what happens if the client closes the connection around when this is called.

PR #191, opened 2025-12-03
AAP-47956 Consistently chunk messages for pg_notify as well as socket brokers

AlanCoding:
We are kind of stumbling into a major architectural change here, so let's recap with the history:

 - socket broker was implemented to get _more stable and reliable_ debugging interaction with the local dispatcherd service on a single node
 - In spite of how reliable this _should_ have been, we have PermissionDenied error trying to connect to the socket which just leaves debugging commands just kind of not working

Even if we could fix that, having unreliable debugging is the **opposite** of what we want for debugging. I don't yet know how this will be adopted in AWX, but the principle of debugging here is **never not work**. In the process of developing the socket communication, we accidentally solved the problem we needed solved for pg_notify anyway.

 - there is a risk that a large message body won't be send as a single packet over the socket, so we used a size small enough to assure this will never happen, like 1024 bytes which for OS-level reasons has assurances it will never be broken up into smaller things.

So this fix is to take what was already developed and apply it to pg-notify as well, so that _if someone does_ run debugging commands over pg_notify, it won't not work.

The bottom line principle here is: Make control-and-reply commands work. Don't care about the details, just make it work.

---

> [!NOTE]
> Introduce JSON chunking for large broker messages, add dispatcher-side reassembly with timeouts, diagnostics, and apply to pg_notify and socket brokers.
> 
> - **Chunking/Core**:
>   - Add `dispatcherd/chunking.py` with `split_message` and `ChunkAccumulator` for JSON chunking/reassembly with timeout and stats.
>   - Wire reassembly into `DispatcherMain` (`process_message`) with stale-chunk expiry; expose accumulator on protocol.
> - **Brokers**:
>   - `pg_notify`: chunk outgoing messages (`apublish_message`/`publish_message`), ensure `UNLISTEN` in `finally`, add `max_message_bytes`, and allow `max_messages: int | None` in `process_notify`.
>   - `socket`: chunk replies and client sends (async/sync), newline-delimited framing, add helper `apublish_message_to_socket`, and support `max_messages: int | None`.
>   - `noop`: type hint update for `process_notify(max_messages: int | None)`.
> - **Control/Diagnostics**:
>   - Control replies now parsed via chunk accumulator; invalid JSON yields `{'error': 'JSON parse error', 'original': ...}`.
>   - New control command `chunks` returns accumulator status/partials.
> - **Service/Config**:
>   - `DispatcherMain` accepts `chunk_message_timeout_seconds`; plumbed via `dispatcher.yml` and `schema.json` (`service.main_kwargs`).
> - **Docs**:
>   - Document chunked message format in `docs/message_formats.md` and add `docs/worker_statuses.md`.
> - **Tests & Demo**:
>   - Add unit/integration tests for chunking, control callbacks, and large pg_notify replies; update `run_demo.py` to showcase chunk diagnostics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 45119c2cf2531ac39478fb83a484c3a9a008c6b8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/dispatcherd/service/control_tasks.py
+++ b/dispatcherd/service/control_tasks.py
@@ -4,7 +4,7 @@ import logging
 
 from ..protocols import DispatcherMain
 
-__all__ = ['running', 'cancel', 'alive', 'aio_tasks', 'workers', 'producers', 'main', 'status', 'partials']
+__all__ = ['running', 'cancel', 'alive', 'aio_tasks', 'workers', 'producers', 'main', 'status', 'chunks']
 
 
 logger = logging.getLogger(__name__)
@@ -137,13 +137,16 @@ async def main(dispatcher: DispatcherMain, data: dict) -> dict:
     """Information about scalar quantities on the main or pool objects"""
     ret = dispatcher.get_status_data()
     ret["pool"] = dispatcher.pool.get_status_data()
-    ret["chunks"] = dispatcher.chunk_accumulator.get_status_data()
     return ret
 
 
-async def partials(dispatcher: DispatcherMain, data: dict) -> dict[str, dict]:
-    """Return partial chunk assemblies keyed by message id."""
-    return await dispatcher.chunk_accumulator.aget_partial_messages()
+async def chunks(dispatcher: DispatcherMain, data: dict) -> dict[str, dict]:
+    """Return chunk accumulator diagnostics and buffered message metadata."""
+    partials = await dispatcher.chunk_accumulator.aget_partial_messages()
+    return {
+        'status': dispatcher.chunk_accumulator.get_status_data(),
+        'partials': partials,
+    }
 
 
 async def status(dispatcher: DispatcherMain, data: dict) -> dict:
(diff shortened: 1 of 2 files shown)


djyasin (approved):

AlanCoding:
Doing some verification with AWX (no changes on AWX side):

```
In [13]: ctl.control_with_reply('chunks', data={'b': 'a'*10000})
2025-12-10 18:35:41,746 INFO     [-] awx.main.dispatch.control control-and-reply message returned in 0.00404810905456543 seconds
Out[13]: 
[{'status': {'total_chunks_received': 60,
   'successful_assemblies': 5,
   'errored_assemblies': 0,
   'timed_out_assemblies': 0,
   'active_messages': 0},
  'partials': {},
  'node_id': 'awx-1'}]
```

Oh I think that's chunking with the socket broker limit, which is 1024, which explains why the number of chunks increases by >10 every time. This should be fine, actually relatively good demonstration of its operation.

PR #199, opened 2025-12-11
Move to per-worker locks

AlanCoding:
Includes or replaces https://github.com/ansible/dispatcherd/pull/181

Experimental as of opening.


AlanCoding:
current mood: not feeling good about this.

Issue #73, opened 2025-02-12
Do some minor linting of test files


AlanCoding:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/dispatcher/issues) for duplicates.
- [x] I understand that dispatcher is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

I saw we had an unused import in a test file, and this is a good example of something we _would_ like to enforce with automated checks for the content in `tests/` folder.

I'm fairly sure we _don't_ want to run MyPY against the test files. I would prefer we don't just apply the rules for the code to the tests, because code quality is not important for the same reasons.

### Steps to reproduce

N/A

### Sugested feature result

title

PR #201, opened 2025-12-17
AAP-56913 Split tests into 2 folders to separate asyncio tests and run linters

AlanCoding:
Fixes https://github.com/ansible/dispatcherd/issues/73

This is a relatively large and mechanical change.

See the `.github/` folder, and you can see how we already were having to run the tests in 2 phases. This is because the pytest-asyncio library creates its own event loop, and we have no control over this. That isn't necessarily exclusive with the other subprocess-based tests but because we don't have control due to the library, it's basically unfixable. The solution is to not mix the subprocess & asyncio tests. It's ambiguous where some tests (like ordinary synchronous unit tests) might go, but it's not actually important. What's important is that there are 2 separate `pytest` invocations.

This also enables linting for test files, which we might as well do along with this change so we just land one big code diff at once.

---

> [!NOTE]
> Separates asyncio tests into `asyncio_tests/`, updates CI/Makefile/pytest config to run sync and async test suites with coverage, and enables linting/type settings for tests; adds/moves related async tests and minor fixes.
> 
> - **Tests**:
>   - Create dedicated async test suite under `asyncio_tests/` (integration + unit), adding async-specific tests (metrics, sockets, pool, control cleanup, connection saver, chunking, broker callbacks, etc.).
>   - Remove async tests from `tests/` and keep only sync tests; minor test cleanups and reorderings.
> - **CI**:
>   - Run two pytest passes with coverage: `tests/` (sync) and `asyncio_tests/` (async) in `.github/workflows/ci.yml`.
> - **Tooling/Config**:
>   - Makefile: add `ASYNC_TEST_DIRS`, run both suites in `make test`; expand linters to include `tests` and `asyncio_tests` via `PYTHON_LINT_PATHS`.
>   - Pytest config (`pyproject.toml`): set `testpaths` to `tests` and `asyncio_tests`.
>   - Mypy: ignore errors for `tests.*`, `asyncio_tests.*`, and `conftest`.
> - **Docs**:
>   - README: document running both `pytest tests/` and `pytest asyncio_tests/`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 072ca03cc0ba3a2a8c0bba83d3a74950fc5d3869. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/asyncio_tests/integration/test_metrics_use.py
+++ b/asyncio_tests/integration/test_metrics_use.py
@@ -74,7 +74,7 @@ async def test_metrics_invalid_utf8_returns_400(ametrics_dispatcher):
     reader, writer = await asyncio.open_connection("localhost", TEST_METRICS_PORT)
     response_bytes = b""
     try:
-        bad_request = b"GE\xffT /metrics HTTP/1.1\r\n" b"Host: localhost\r\n" b"Connection: close\r\n" b"\r\n"
+        bad_request = b"GE\xffT /metrics HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n"
         writer.write(bad_request)
         await writer.drain()
 


Alex-Izquierdo (approved):
My only concern is about good practices and proper history record. This PR in the end does not only split the tests, also removes some of them and applies some format changes. I suggest to split it in multiple PR's to have one commit/PR per change and have a more accurate a meaningful git history, or at least update the title/description to cover all the changes included. 

Regardless of that LGTM

PR #219, opened 2026-02-24
AAP-66379 Fix failure to scale-down bug and increase scale-down speed

AlanCoding:
We saw excessive memory use, and this seems to be caused by a race condition where we get a lot of tasks, and some tasks finish while we're still in the code path of dispatching other tasks. This leaves the dangerous internal data structure of last-used-by-counts filled to an unexpected worker count, and the method conflated a discovered `None` value with that meaning that we _need_ that many workers.

---

> [!NOTE]
> **Medium Risk**
> Changes core worker auto-scaling/retirement logic and introduces new state tracking under concurrency; mistakes could cause over/under-scaling or delayed worker retirement. Added tests and lock-scoped updates mitigate risk, but behavior changes are significant in production workload patterns.
> 
> **Overview**
> Fixes scale-down race/incorrectness by replacing `last_used_by_ct` with a demand-aware `WorkerUsageTracker` that distinguishes *blocked* (at/below demand) from idle timestamps, updates on task completion (`record_task_finish`), and provides debug summaries in `get_status_data()`.
> 
> Updates `WorkerPool.scale_workers()` to retire surplus workers in a loop (down to `min_workers` and `demand + scaledown_reserve`), capped at 300 stops per tick, and adds new `scaledown_reserve` configuration to `schema.json`.
> 
> Adds extensive new tests for scaledown behavior and diagnostics, plus a fake pool/process manager test harness to avoid real multiprocessing where not needed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 949edaf5a43aaab4be6ee184f56b244596a27460. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/asyncio_tests/unit/service/conftest.py
+++ b/asyncio_tests/unit/service/conftest.py
@@ -1,6 +1,5 @@
 import queue
 from typing import Callable
-from unittest import mock
 
 import pytest
 
--- a/asyncio_tests/unit/service/conftest.py
+++ b/asyncio_tests/unit/service/conftest.py
@@ -8,6 +8,14 @@ from dispatcherd.service.pool import WorkerPool
 from dispatcherd.service.process import ProcessManager
 
 
+def fill_and_check_scale_down(pool: WorkerPool) -> bool:
+    """Test helper: fill the usage tracker with current pool state and check if scale-down is warranted."""
+    worker_ct = len([w for w in pool.workers if w.counts_for_capacity])
+    demand_ct = pool.active_task_ct()
+    pool.usage_tracker.fill_unknown_usage(worker_ct, demand_ct)
+    return pool.usage_tracker.should_scale_down(worker_ct, demand_ct)
+
+
 class FakeProcess:
     """A stand-in for ProcessProxy that requires no multiprocessing resources.
 
--- a/asyncio_tests/unit/service/conftest.py
+++ b/asyncio_tests/unit/service/conftest.py
@@ -1,4 +1,6 @@
+import queue
 from typing import Callable
+from unittest import mock
 
 import pytest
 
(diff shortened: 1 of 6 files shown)


asyncio_tests/unit/service/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+@pytest.fixture
+def pool_factory(test_settings) -> Callable[..., WorkerPool]:
+    def _factory(**kwargs_overrides) -> WorkerPool:
+        pm = ProcessManager(settings=test_settings)
+        kwargs = dict(process_manager=pm, min_workers=5, max_workers=5, shared=SharedAsyncObjects())
+        kwargs.update(kwargs_overrides)
+        pool = WorkerPool(**kwargs)
+        return pool
+
+    return _factory
```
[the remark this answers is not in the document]
AlanCoding: Hard disagree. He or she who calls `start_working` has the obligation to call `shutdown`.

This is expected to commonly be used in unit testing without the tasks started.

dispatcherd/service/pool.py
```diff
@@ -663,7 +716,7 @@ async def process_finished(self, worker: PoolWorker, message: dict) -> None:
         logger.debug(msg)
 
         running_ct = self.get_running_count()
-        self.last_used_by_ct[running_ct] = time.monotonic()  # scale down may be allowed, clock starting now
+        self.usage_tracker.record_task_finish(running_ct)
```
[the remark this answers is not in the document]
AlanCoding: I don't think so? Because a task finished. The current time is the last time that we used X number of workers, where X includes the current worker. After we subtract 1, we are at a usage level that we are _still using_

dispatcherd/service/pool.py
```diff
@@ -229,6 +229,104 @@ def move_to_end(self, worker_id: int) -> None:
...
+        return result
+
+    def _top5_summary(self) -> dict[int, str | float]:
+        """Return the 5 highest worker-count entries from _last_used_by_ct, with human-readable values."""
+        return self._format_entries(sorted(self._last_used_by_ct, reverse=True)[:5])
+
+    def _near_worker_ct_summary(self, worker_ct: int) -> dict[int, str | float]:
+        """Return _last_used_by_ct values for keys around the current worker count."""
+        keys = list(range(max(worker_ct - 2, 0), worker_ct + 3))
+        return self._format_entries(keys)
```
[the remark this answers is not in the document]
AlanCoding: Addressed in recent push.

dispatcherd/service/pool.py
```diff
@@ -390,19 +467,42 @@ async def scale_workers(self) -> int:
...
+    async def _scale_down(self, prior_worker_ct: int) -> int:
+        """Retire idle workers down to demand + reserve, respecting min_workers.
+
+        Returns a negative count of workers that were sent stop signals.
+        Must not be called if worker_ct <= min_workers.
+        """
+        max_scaledown_per_tick = 300
+        changed_ct = 0
+        async with self.workers.management_lock:
+            # fill_unknown_usage mutates usage_tracker — must be under lock (see also process_finished)
```
fosterseth, line 484: _do_ any of our tests cover this management_lock contention for the 2 places we attempt to update usage_tracker data (i.e. process_finished and fill_unknown_usage)

claude code spit out this test

```
@pytest.mark.asyncio
async def test_fill_blocks_scale_down_when_task_done_but_not_yet_processed(fake_pool_factory):
    """Simulate a task that has finished in the subprocess but whose
    process_finished() call is blocked waiting for management_lock.

    Because mark_finished_task() has not yet cleared current_task, the worker
    still appears busy.  fill_unknown_usage() should see this as active demand
    and block scale-down — the conservative-correct behavior that prevents
    premature worker retirement."""
    pool = fake_pool_factory(min_workers=1, max_workers=10)

    workers = []
    for _ in range(3):
        worker_id = await pool.up()
        workers.append(pool.workers.get_by_id(worker_id))

    # All 3 workers are running tasks
    for i, w in enumerate(workers):
        await w.start_task({'task': f'busy-{i}'})

    base = 1000.0
    # First tick: all 3 keys blocked
    with patch('time.monotonic', return_value=base):
        assert fill_and_check_scale_down(pool) is False

    # Worker 2's task has finished in the subprocess, but process_finished()
    # hasn't run yet — current_task is still set.  This is the state when
    # _scale_down() holds management_lock and process_finished() is waiting.
    # Do NOT clear current_task — that's the whole point of this test.

    # After scaledown_wait, fill_unknown_usage should still see 3 active tasks
    # and block scale-down because current_task hasn't been cleared.
    with patch('time.monotonic', return_value=base + 100.0):
        assert fill_and_check_scale_down(pool) is False

    # Now simulate process_finished() finally running: clear the task and record
    workers[2].mark_finished_task()
    with patch('time.monotonic', return_value=base + 100.0):
        pool.usage_tracker.record_task_finish(3)

    # Now demand is 2, capacity is 3 — scale-down should be allowed
    with patch('time.monotonic', return_value=base + 100.0):
        assert fill_and_check_scale_down(pool) is True
```

but maybe you find that unnecessary

fosterseth (approved):
great changes to make the code more robust

dispatcherd/service/pool.py
```diff
@@ -609,8 +709,6 @@ def active_task_ct(self) -> int:
     async def post_task_start(self, message: dict) -> None:
         if 'timeout' in message:
             await self.timeout_runner.kick()  # kick timeout task to set wakeup
-        running_ct = self.get_running_count()
-        self.last_used_by_ct[running_ct] = None  # block scale down of this amount
```
AlanCoding, line 613: Just to say it out loud, I'm really happy to get rid of this call.

I had a lot of analysis with claude for this, and we have 2 hooks available, the task pre/post, and the periodic task that runs to actually do the scale-down. The task post-hook _really is_ necessary to do scale-down without an additional tick, but post-task is not necessarily performance-sensitive, and it is less clear here.

But basically, the periodic task doesn't need you to tell it the usage count. Because the periodic task knows the usage at the time that matters, which is when _that_ code matters. So marking the blocked values (everything below current usage), is far better to be done in the periodic task.

dispatcherd/service/pool.py
AlanCoding: Does seem to fail on this?

```
        # Now demand is 2, capacity is 3 — scale-down should be allowed
        with patch('time.monotonic', return_value=base + 100.0):
>           assert fill_and_check_scale_down(pool) is True
E           assert False is True
E            +  where False = fill_and_check_scale_down(<asyncio_tests.unit.service.conftest.FakeWorkerPool object at 0x7f7545491cd0>)
```

_thinking_

AlanCoding: passes when I change it to

```python
    # Now demand is 2, capacity is 3 — scale-down should be allowed
    with patch('time.monotonic', return_value=base + 200.0):
        assert fill_and_check_scale_down(pool) is True
```

You ran the check at `base + 100.0` which is the same time at which the task finished. So to rephrase...

The dispatcher was running with 3 tasks and 3 workers. Then 1 worker finished 1 task. It marks it as finished. _Currently_, right after that event, the scale-down decision is "no", do not scale down. After some time passes (100 for what I did), the scale down decision will change to "yes". That's because of the scale-down grace period. After processing a burst, we have to wait until letting go of the capacity, from a be-on-your-toes principle.

AlanCoding: I don't particularly dislike the test, but I think I'll go ahead without it.

It's hard to cover higher level things like "you should always use the lock for these things". But higher level functional behavior would be good to test more of. An aggressive integration test might be the direction I go, customizing the config for the scaledown_wait to be like 1/10 of a second.

PR #220, opened 2026-03-02
feat: enforce uuid4 format and add origin field to message data

### LINKED ISSUE #96: Enforce UUID format, add "origin" information in message data

VedantMadane:
Fixes #96. This PR adds validation to ensure that all message UUIDs are in valid UUID4 format, and introduces an 'origin' field to help identify the source of messages.

### DIFF
--- a/dispatcherd/registry.py
+++ b/dispatcherd/registry.py
@@ -1,6 +1,5 @@
 import inspect
 import json
-import logging
 import threading
 import time
 from typing import Callable, Iterable, Optional, Set, Tuple
@@ -10,8 +9,6 @@ from .config import LazySettings
 from .config import settings as global_settings
 from .protocols import ProcessorParams
 from .utils import MODULE_METHOD_DELIMITER, DispatcherCallable, resolve_callable
-
-logger = logging.getLogger(__name__)
 
 
 class DispatcherError(RuntimeError):
@@ -105,8 +102,7 @@ class DispatcherMethod:
                 if str(val) != uuid:
                     raise ValueError
             except (ValueError, TypeError, AttributeError):
-                logger.warning(f"Invalid UUID4 format provided: {uuid}. Replacing with self-generated UUID4.")
-                uuid = str(uuid4())
+                raise ValueError(f"Invalid UUID4 format provided: {uuid!r}. Caller must supply a valid uuid4 string.")
         else:
             uuid = str(uuid4())
 
--- a/dispatcherd/registry.py
+++ b/dispatcherd/registry.py
@@ -4,7 +4,7 @@ import logging
 import threading
 import time
 from typing import Callable, Iterable, Optional, Set, Tuple
-from uuid import uuid4
+from uuid import UUID, uuid4
 
 from .config import LazySettings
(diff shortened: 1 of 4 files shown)


dispatcherd/registry.py
```diff
@@ -97,8 +98,22 @@ def get_async_body(
...
-        body['uuid'] = uuid or str(uuid4())
+
+        if uuid:
+            try:
+                val = UUID(uuid, version=4)
+                if str(val) != uuid:
+                    raise ValueError
+            except (ValueError, TypeError, AttributeError):
+                logger.warning(f"Invalid UUID4 format provided: {uuid}. Replacing with self-generated UUID4.")
+                uuid = str(uuid4())
```
AlanCoding: In _most_ cases it would be less bug-prone to have this throw an exception. We do provide this in AWX, and that should be fine and compatible with the validation, but if it gets rejected, tracking down what happened will be really hard.

VedantMadane: Good point -- updated to raise `ValueError` instead of silently replacing. Invalid uuid4 inputs now fail loudly.

VedantMadane: Good point. The `str(val) != uuid` normalization check catches case differences and non-canonical formatting, but the error message already includes the original value via `{uuid!r}` so the caller can see exactly what was rejected. AWX uses `uuid4()` output which is always lowercase canonical form, so it should pass cleanly. If this ever becomes a friction point we could downgrade the strict equality check to a warning, but for now I think raising is the safer default.

VedantMadane: Agreed — invalid uuid4 input continues to raise `ValueError` (no silent replacement). Tightened the validation path and added unit tests covering:

- malformed / non-uuid strings
- non-canonical uppercase form
- canonical uuid4 acceptance
- auto-generation when omitted

AWX's `uuid4()` output is canonical lowercase and passes cleanly.

32705897e1534f219949c35876c83ce4df42e770

PR #222, opened 2026-05-21
fix: false-positive health checks from stale pgnotify replies

ttuffin:
Notifications from previous LISTEN sessions can be picked up by subsequent health checks, which causes workers to appear healthy when they are not. This fix includes channel validation for named channels, to discard replies that arrive on a different channel and preemptively UNLISTEN before LISTEN in process_notify to reduce stale notification buildup.

I have modified the existing test to account for the new channel validation and added a new test to validate stale notifications are discarded.

Assisted-by: Claude

https://redhat.atlassian.net/browse/AAP-76207

### DIFF
--- a/dispatcherd/brokers/pg_notify.py
+++ b/dispatcherd/brokers/pg_notify.py
@@ -352,6 +352,8 @@ class Broker(BrokerProtocol):
 
         with connection.cursor() as cur:
             try:
+                # Clear any stale LISTEN state from previous users of this cached connection
+                cur.execute(self.get_unlisten_query())
                 for channel in self.channels:
                     cur.execute(self.get_listen_query(channel))
                     logger.info(f"Set up pg_notify listening on channel '{channel}'")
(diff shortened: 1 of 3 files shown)


dispatcherd/brokers/pg_notify.py
```diff
@@ -352,6 +352,8 @@ def process_notify(self, connected_callback: Callable | None = None, timeout: fl
 
         with connection.cursor() as cur:
             try:
+                # Clear any stale LISTEN state from previous users of this cached connection
+                cur.execute(self.get_unlisten_query())
```
AlanCoding, line 356: Had to think about it, but I guess this doesn't cause any race conditions for control-with-reply because `connected_callback` is fired after this line.

AlanCoding (approved):
Interesting, makes sense. The synchronous `process_notify` is pretty specific to this purpose, so there's not any good argument against it.

